PR #19, opened 2020-08-19
Update Unity Bootstrap

nathan-rollins:
This PR updates the ASU Unity Bootstrap resources and CSS. Minor CSS updates for Cards and mobile accessibility are included. This update also includes the latest mobile and accessibility tweaks for the ASU Header.


nathan-rollins:
Good points.

* Let's verify with Stetson about the Parent Unit link.
* Thanks for pointing out the variable names. A legacy of the original code from WP1.0 theme. I have pushed up an update for this.

nathan-rollins:
I'm going to merge this in now. IF for some reason, we need to change the behavior of the Parent Unit link, we can do that in a separate PR.

PR #21, opened 2020-09-14
Add Highlight Options to Hero Title

wmcconnell:
Add the ability for users to select a highlight style for the hero image title text:

* Added a new field to the existing ACF field group that allows the user to select "gold", "black", or "none" as a text highlight for hero images

* Added corresponding code in the `hero.php` template partial to implement the new option.

* Cleaned up the existing `hero.php` to remove some unnecessary CSS classes, and some inline CSS that would conflict with our current plans to limit hero images to a maximum of 1920px

### DIFF
--- a/acf-json/group_5ef954da477f5.json
+++ b/acf-json/group_5ef954da477f5.json
@@ -67,6 +67,31 @@
             "maxlength": ""
         },
         {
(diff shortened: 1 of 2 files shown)


nathan-rollins (approved):
Looks good to me. Good job. :)

PR #22, opened 2020-10-01
Child theme compatibility - get_template_directory vs get_stylesheet_directory

SteveRyan-ASU:
The use of `get_stylesheet` functions in various parts of the parent theme will fail when in the presence of a child theme. These are a set of corrections that should force the parent theme to look within ONLY the parent theme for the various called resources.


nathan-rollins (approved):
Excellent update, thank you!

PR #23, opened 2020-10-01
A proposed name change for the parent theme - ASU TwentyTwenty

SteveRyan-ASU:
A humble suggestion. The theme name is fairly long and doesn't show up well in `/themes.php`. 

In looking for a replacement name, I reached for the obvious and added ASU in the front of it. The implication being that this is ASU's "generic" or "starter" theme for WordPress.

It eliminates the version (2.0) from the name as well the mention of "web standards" in the title. But, I think it still gets its point across.


SteveRyan-ASU:
image

...comes with a PSD that makes incrementing the version number easy. :-)

nathan-rollins:
I don't know, this makes it look like this theme is connected with the TwentyTwenty theme bundled with WP. WP2020 was just a random idea I had anyway. What if we just call this "ASU WS2.0"? And thank you for the improved thumbnail!

nathan-rollins:
Ok, I'm going to close this PR now that we've settled on our new project name.

PR #27, opened 2020-10-20
Navigation menu fixes.

SteveRyan-ASU:
This pull request addresses two bugs in the menu creation process.

First: the construction of a mega menu should be triggered where there are greater than 2 sets of grandchildren.

- Initial logic was set to be > 3 in accordance with an early UDS design spec.
- Changed the number to be > 2 and we're back in business.

Second, the example menu content was being recreated at every instance if `after_theme_setup` which was basically every time an admin page was loaded. This can result in a problem by which an alternate menu object in the DB can become unassigned from any of the three menu regions. 

- Moved the content creation steps out of the `after_theme_setup` hook.
- Created a new function to run at `after_switch_theme` which is the hook that runs one time after the theme is activated. 
- Removed the logic that actually assigns the example menus to a menu location.


wmcconnell (approved):
Looks good, Steve. I like the cleanliness of moving the theme activation stuff to its own place. Mega menu worked as advertised on my local site, and we've got the new theme action in place. Approving.

wmcconnell:
Steve updated the code to use the `after_switch_theme` hook. All checks have passed. Merging.

PR #33, opened 2020-10-30
Prevent Travis from Updating Composer

### LINKED ISSUE #32: Travis Builds Failing
Travis CI Builds are currently failing on particular versions of PHP because Travis is using the latest version of Composer (2.0) and not all our dependencies are ready for that. We'll need to temporarily configure Travis to stick with the latest 1.x version of Composer

wmcconnell:
Builds are failing on certain (older) versions of PHP during our Travis CI testing. This addition to the `travis.yml` file will prevent Travis from automatically updating to Composer 2.0 when it creates our test environments.

This fixes #32

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,12 @@ matrix:
   allow_failures:
     - php: nightly
 
+# Control installation settings
+before_install:
+  # prevent composer from automatically updating to latest
+  # (until we are ready for Composer 2.0)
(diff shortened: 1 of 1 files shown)


wmcconnell:
All checks passed (which was what this PR was trying to fix). Going ahead with admin power to merge this so that other PR's and local branches can pull latest to avoid the same build errors.

Issue #31, opened 2020-10-30
ACF: Revisit our decision to include ACF within the theme


SteveRyan-ASU:
The decision to include the free version of ACF wiithin our theme was made in order to take advantage of the plugin's capabilities (as developers) while not requiring additional actions for our users. 

That decision should be revisited in light of two important conversations that happened during our last UDS-WordPress project meeting:

- Including the plugin files within the theme requires the development team to manually update the plugin whenever an update is issued from the ACF team
- In order to take advantage of ACF Pro, the original bundled theme files need to be first "unplugged" from the theme.

SteveRyan-ASU:
Offline @nathan-rollins and I chatted and figured out that the progression forward here would involve testing and discovery of a couple of additional items.

1. We would want to know whether or not the bundled ACF JSON files would still work if ACF was no longer directly included with the theme. 
2. We would want to understand how to best go about recommending (or even requiring) our end users to install at least the free version of ACF from the normal WordPress repository.

SteveRyan-ASU:
Testing the first bullet point above. Commenting out the following line in `functions.php` effectively prevents the theme from loading the included version. https://github.com/asu-ke-web-services/UDS-WordPress-Theme/blob/bd28bd3b84bf5256fa64899c292eeba0f1418213/functions.php#L27

Then, when installing the ACF plugin from the repo and looking at the admin menu, I verified that the JSON files for our created ACF fields were present and waiting to be synced from the DB. The same is also true when activating the ACF Pro version of the plugin.

The UI for ACF also provided a clue that the plugin (out of the box) knew to look for config files in the `/acf-json` subfolder of the theme.

![image](https://user-images.githubusercontent.com/2085753/97761062-f799e380-1ac1-11eb-9c7d-214749b1ef9a.png)

SteveRyan-ASU:
For the second bullet point above, I called upon my mental Rolodex of WordPress utility widgets. I remembered hearing about a PHP library for WordPress that provides a mechanism for recommending or installing other complementary plugins & themes. It was this one:

- https://github.com/TGMPA/TGM-Plugin-Activation
- https://tgmpluginactivation.com/

This library is well documented. It makes the claim that this is the recommended way that products sold within Code Canyon as well as distributed via WordPress.org should recommend other things to be installed. The idea here would be to configure this library to work with our theme, and then, upon initial activation:

- recommend or require ACF from the WordPress.org repo.
[...]

PR #35, opened 2020-11-02
Use TGMPA to require Advanced Custom Fields

SteveRyan-ASU:
This PR is related to Issue #31 and would resolve all issues presented there.

The TGMPA library is a PHP class that you can include in your theme that handles the recommendation of plugins to work with your product in a nice "looks like WP native" kind of way. As configured here, it requires the use of Advanced Custom Fields, either as the version downloaded from the regular WP plugin repository, or as a self-installed PRO version of the same plugin.

Notes:
- The action I chose to look for to allow for multiple versions of the same plugin (normal vs. pro) to fulfill the is_active requirement of TGMPA was the `get_field` action from ACF. 
- Currently, ACF is a _recommended_ plugin only. You could get away with never building a site without CTA buttons in the header.
- Functionally, there isn't too much of a difference in a recommended vs a required plugin other than the language used in the warning.


functions.php
```diff
@@ -24,11 +24,11 @@
 	'/custom-comments.php',                 // Custom Comments file.
 	'/jetpack.php',                         // Load Jetpack compatibility file.
 	'/wp-custom-fields.php',                // Disable WP Core custom fields metaboxes.
-	'/advanced-custom-fields.php',          // Load integrated ACF plugin.
 	'/class-wp-social-media-walker.php',    // Load custom WordPress nav walker for footer Social Media menu wdiget.
 	'/wp-custom-menu.php',                  // Load custom menu builder functions.
 	'/editor.php',                          // Load Editor functions.
 	'/deprecated.php',                      // Load deprecated functions.
+	'/tgm-plugin-activation.php',            // Load TGM Plugin Activation script for dependent plugin recommendations.
```
nathan-rollins, line 31: The order of these files has some significance. There's no hard rules necessarily, except that some later include files depend on functions being already loaded prior to them. And baseline setup of the theme occurs early in the order. I would suggest moving the tgm activation line up in the order, and since it is effectively replacing the old acf line, to move it into the same position. Unless there is a specific reason to load it late. :)

SteveRyan-ASU:
The plugin presents a dialogue to the user when the theme is first activated to install ACF from the WP repo.

- The **Begin activating plugin** link takes you through another dialogue to do exactly that. A few steps later, after installing the plugin (if it's not already installed), you're good to go.
- Activating either ACF or ACF pro right from the plugin screen makes the warning go away.
- Pressing the X button in the upper right-hand corner of the banner removes it from the current screen. But it reappears when you navigate to the next admin screen.
- Clicking the **Dismiss this notice** link has the effect of removing the banner permanently. 

image

SteveRyan-ASU:
Other nice to know things:

- If the warning is dismissed entirely, there is still a place in the admin area that the user can go to in order to see what plugins are missing or uninstalled.
- The dialogue is great about telling you what you are installing and potentially FROM WHERE you are installing it as well. The plugin works to bundle code together from all sources - Github, the WP repo, included plugin files with the theme or other arbitrary places as well. 

image

nathan-rollins (changes requested):
Looks good. Just a small suggested change in the functions.php.

Are we happy with the default TGM settings? I presume we don't want to force the notice to be undismissable or to automatically activate the plugin on install, just in case that interferes with installing the Pro plugin as an alternative.

functions.php
SteveRyan-ASU: Good call. That makes sense to me. :-)

nathan-rollins: Although, if installing the Pro plugin does clear the notice, then maybe we should make the notice undismissable.

SteveRyan-ASU:
One last point of discussion here is about the way that this is installed to the theme. The library outlines one of two ways - either directly copying the class from the library and dropping it into your code, or as a Composer dependency.

I went with option 1 here only because it didn't look like there were ANY other composer pieces that were required to be able to use the theme properly as of right now, and I didn't want to create that burden for shipping versions of the theme out to our end users unless we had to. 

...puts the burden back on us to monitor the TGMPA code for updates. (Which kinda defeats the purpose of unbundling ACF, right? 😄  ) But, it would scale better if we happened to need or want to require additional plugins.

functions.php
SteveRyan-ASU: Added code to move the declaration up to the previous spot where ACF was loaded.

SteveRyan-ASU: Yes, I think that the default choices are good ones for now. 

- Depending on what other things we ship via this mechanism, we may want to revisit the `dismissable` setting to bug our users more. This would be particularly true if we wanted to require something like our own block library plugin along with this theme - if it woudn't work out of the box on some level, then we can tighten things up a bit there.
- The auto-activate on install is one less click, but I think requiring that click from the end-user solidifies the experience of installing/activating a plugin as a "normal" behavior. 👍

nathan-rollins:
Ah, Houston, we may have a problem.

TGM Plugin Activation hasn't been updated since May 2016. Its Issues queue has a massive backlog that hasn't been addressed.

Also of concern is the 3.0 Roadmap which states goals to fix:

* Multisite compatibility
* Fix conflicting UI messages for usage in themes / plugins
* Improved UI by having the admin page always in the same location
* Updatability of TGMPA independently of the plugin/theme which ships it
* Show dependencies more clearly

There are multiple bug reports regarding functionality of this plugin in multisite environments and newer versions of WP. Some of this may be due to self-inflicted installation errors by devs, but we need to be thorough in testing this PR before we accept it.

nathan-rollins:
Here's the GitHub Milestone page for 3.0: https://github.com/TGMPA/TGM-Plugin-Activation/milestone/2

Note that there are several issues tagged for multisites. One indicates they are not able to update their theme while using TGM.

nathan-rollins:
As a counterpoint, according to this discussion thread, as of April 2019, the project maintainer considers the plugin to be fine with only a few minor todos in his queue: https://github.com/TGMPA/TGM-Plugin-Activation/issues/741

So, as long as we are able to validate this branch works fine in multisite mode and WP 5.5, then maybe everything is fine? The other multisite discussion I observed concerned the TGM notifications not showing up at the Network admin level. They were only visible on each subsite dashboard. That is less than ideal, but maybe we can work around it with proper documentation.

SteveRyan-ASU:
Copy. I'll see if I can document what the experience is with these code modifications against a multisite installation. I can spin one up easily enough with WP Docker. 😁

wmcconnell:
Just wanted to confirm expected behavior before I approved anything:

- I pulled down Steve's branch and then deactivated my ACF PRO plugin
- I saw the dashboard message box letting me know that ACF was suggested
- I dismissed that dialog box, and it will not reappear no matter what I do

As you mentioned, the `Appearance/Install Plugins` area does show me ACF (non-pro version), but the title of that page is 'Install Required Plugins'.

wmcconnell:
I'm trying to figure out the issue that is preventing one of the builds from working. It's complaining that the lock file is not up to date with `composer.json`. I noticed that Carbon Fields is still referenced in the `composer.lock` file, but has been removed from `composer.json`. Could that be it?

As if to answer my own question: I ran `composer update` on my local copy of Steve's branch and it removed Carbon Fields, while also updating PHPCS and `roave/security-advisories`. Running `composer validate --no-check-all --strict` (as Travis CI does on the failed build) no longer generates the error message.

wmcconnell (changes requested):
This seems to work as expected, but develop has moved on since the branch was made, and we still have errors on Travis. I think the following two things should address all that:

- Pull in the latest from develop
- Run `composer update` to fix the composer issues popping up in Travis.

If we do that and we don't get any issues in Travis, I'm ready to approve.

SteveRyan-ASU:
Addressed lingering concerns about merging with the `develop` branch and composer. Should be good to go. 👍

nathan-rollins (approved):

wmcconnell (approved):

Issue #54, opened 2020-11-10
Action hooks for easier child theme templating


SteveRyan-ASU:
When adding templates to a child theme of **UDS-WordPress**, it will be helpful to have a series of defined places within the parent theme upon which to add code. Within the WordPress ecosystem, this is typically done by the simple insertion of a "hook" in a documented place within the theme or plugin. Child themes or additional plugins can then take advantage of this prescribed architecture by inserting a template or custom piece of code in a safe and secure manner. 

Action hook overview: https://kinsta.com/blog/wordpress-hooks/ (Long read.)

I wanted to raise this issue as a way to introduce the topic to the group and to think through a way to:

- name the hooks correctly
- talk briefly about the risk (or lack thereof) that inserting these pose to the end product
- figure out how to document their existence someplace.

SteveRyan-ASU:
For the project that I am working on right now, I can foresee a use case for two such "hooks."

- **Hook 1: After Header** - fires immediately after the closing `</header><!-- end #asu-header -->` statement in `header.php` That would allow for a small banner or other alert mechanism to be added before a potential hero image across multuple pages on the site. 
- **Hook 2: Before Footer** - fires immediately before the closing global `<footer>` tag in `footer.php` This would allow for a persistent "block" of code to appear across multiple pages as a closing feature for a site.

In addition, I would like to raise the possibility of a third hook, but perhaps leaving it undocumented for the moment.
- **Hook 3: Before Footer Columns** - would fire immediately before the `<div id="wrapper-footer-columns">` also in `footer.php`. In conjunction with the theme option to turn off the native footer column feature, this would be a handy way to "roll your own footer columns" either using widgets or other means.

SteveRyan-ASU:
Adding a hook to the parent theme should have exactly zero effect on the output when used as a standalone theme.

SteveRyan-ASU:
Example from the current furi.engineering.asu.edu site.
The hook in the header would allow for the creation of a feature like this one.
![image](https://user-images.githubusercontent.com/2085753/98631633-1296ff80-22db-11eb-8f55-15dd3e8e6b29.png)

... and the hook above the footer would allow for the insertion of a feature like this near the footer.
![image](https://user-images.githubusercontent.com/2085753/98631691-2fcbce00-22db-11eb-9b9c-169100f81376.png)

SteveRyan-ASU:
[...]

PR #55, opened 2020-11-10
Add hooks: after_header, before_footer and before_footer_columns

SteveRyan-ASU:
Proof of concept for the issue raised in #54 a moment ago. The following three lines of code should have no outward effect on the parent theme itself. However, within a child theme, that would allow for a developer to do something like the following in the child theme's `functions.php` 
```
/** 
 * Adds a section of content right above the global footer.
 */
function uds_furi_add_symposium_totals() {
	get_template_part( 'templates/snapshot', 'footer' );
}
add_action( 'uds_wp_before_global_footer', 'uds_furi_add_symposium_totals' );
``` 

... which would look for `/templates/snapshot-footer.php` and execute whatever that code might be immediately before outputting the footer.


SteveRyan-ASU:
I'll mention the alternative here would be for me to copy the existing `footer.php` from the parent theme into the child theme and add the `get_template_part()` call in the correct place. 

The obvious benefit to having a hook available is that the hook allows for the parent's `footer.php` to be updated without the need to also update the code in the child theme.

nathan-rollins (changes requested):
I think this is a fantastic idea, and these are great examples for the rest of the team to keep in mind for extension opportunities elsewhere in the theme.

That said, I do have a request: please write up documentation notes in the README for this.

We need to improve our new feature pipeline to include proper documentation in the README at the time of new feature additions. We should actually make this a requirement of all our new feature PRs, they can't be closed out without appropriate documentation for end users or devs being included.

It is important that we include details about why hooks are included, and what each hook allows a theme developer to do (maybe a table of function names with file and line number and brief description?)

nathan-rollins (approved):

nathan-rollins:
I am assuming this PR is ready for re-review, since the README is updated and looks pretty good. Nice job!

PR #65, opened 2020-11-24
Zalsidiki sidebar margin

zainabD:
I wrapped the sidebar area with `<div class="my-12">`, to add margin to both top and bottom to make the area aline with content area

- **my-12** is a bootstrap class.
- I updated both files: 
> - templates-sidebar/sidebar-left.php 
> - templates-sidebar/sidebar-right.php


nathan-rollins:
Is there a specific reason this margin class is being added in a new div, rather than being added to the main sidebar div?

zainabD:
Yes, some widgets have margin-top value on their titles, If my-12 class was added to the main sidebar div, the additional margin on the widget's title will count. 
We only want the 6rem margin that the class my-12 adds to be there, therefor, adding my-12 in a new div will make the additional margin collapse (check **Margin Collapse** concept in W3 schools). 
- To test what I'm talking about, you can try to place a search widget with a title added on one of the sidebar areas, and from the browser inspect, try to place the my-12 class to the main div classes.

-- I would also hold on on this branch since we may end up replacing the whole sidebar and templates work with Steve's work: [sryan-page-templates](https://github.com/asu-ke-web-services/UDS-WordPress-Theme/pull/72)

wmcconnell (approved):
Approving per our discussion in the Sprint Planning meeting today - with the caveat that we may end up undoing or changing this based on Steve's work. Also, I learned about margin collapse today!

nathan-rollins (approved):

PR #66, opened 2020-11-24
Customize WordPress Widget Menus for ASU Web Standards theming

zainabD:
Extended Walker_Nav_Menu class to make the side menu match the one in Storybook

- Created new file for the widget wrapper: inc/navigation-widget.php
- New file for the Walker_Nav_Menu class extension: inc/class-custom-walker-nav-menu.php
- Updated functions.php, called the file **navigation-widget.php**

> In **inc/navigation-widget.php**:
> - Created function **uds_wp_theme_custom_widget_title_tag** to customized the widget wrapper and its title's wrapper only when the **Navigation Menu** widget is in use.
> - Created function **uds_wp_theme_special_nav_class** to replace the default class 'current-menu-item' when the menu item is selected, replace it with 'active' class. 
> - Created function **uds_wp_theme_navigation_menu_custom_walker** to customize the menu item wrapper and walker. excluded 'social-media' menus, since these have their own walker.

> In **inc/class-custom-walker-nav-menu.php**:
> - extended the class Walker_Nav_Menu to override some of the walker functions: start_lvl, end_lvl and start_el


nathan-rollins (commented):
The new nav walker class is too generic. Based on the name and file comments, it just looks like a basic menu builder that would be used anywhere in the site. Since this is a menu builder specifically for the sidebar, it should be named clearly as a custom sidebar navwalker. And the comments should be clear why we need this class and what it does for us. (Why did you need to create a new navwalker?)

A second question is whether a custom nav walker is the right solution for this feature. We have implemented a custom menu builder system (no Nav Walkers) for the main navigation and footer menus, so I wonder if it wouldn't be better to leverage our menu builder instead?

But I guess that depends on the goal of this PR and the feature you are working on. It isn't clear to me at all what is going on here and why. And there's no documentation in the Readme! I recommend a team review of this PR to see if this approach is what we really need.

zainabD:
> The new nav walker class is too generic. Based on the name and file comments, it just looks like a basic menu builder that would be used anywhere in the site. Since this is a menu builder specifically for the sidebar, it should be named clearly as a custom sidebar navwalker. And the comments should be clear why we need this class and what it does for us. (Why did you need to create a new navwalker?)
> 
> A second question is whether a custom nav walker is the right solution for this feature. We have implemented a custom menu builder system (no Nav Walkers) for the main navigation and footer menus, so I wonder if it wouldn't be better to leverage our menu builder instead?
> 
> But I guess that depends on the goal of this PR and the feature you are working on. It isn't clear to me at all what is going on here and why. And there's no documentation in the Readme! I recommend a team review of this PR to see if this approach is what we really need.


- It is a basic menu builder that would be used anywhere in the site except when social media menu is called.
- I have a description added there says "Description: A custom WordPress nav walker class to customize navigation menu element wrapper and child menu's wrapper."
- Why am I building this: to apply the new standards to nav menus. 

Answering the second question:
- Yes, based on my knowledge, I think that the custom nav walker is the right solution for this feature, since this will customize any menu added to the website, except **main menu**, **footer menu** and **social media menu**. And since we don't know whether this menu will be added to right sidebar, left sidebar or any other sidebar, or even through a plugin in the future, all menus other that these 3 mentioned above should follow the standards, and therefore I added the needed wrappers and classes by extending the Walker_Nav_Menu. (let me know if you have a better solution)

- I didn't know that we need to add a documentation in the Readme file before approval, but I will add that part to make things clear.

nathan-rollins:
Thank you, based on your original PR description, I thought this was a solution only for the side menu nav. I understand now this *is* a universal fix to make all menus Web Standards compliant.

inc/class-custom-walker-nav-menu.php
```diff
@@ -0,0 +1,113 @@
...
+ * Navigation menu walker class
+ *
+ * @package uds-wordpress-theme
+ */
+
+// Exit if accessed directly.
+defined( 'ABSPATH' ) || exit;
+
+/*
+ * Class Name: Custom_Walker_Nav_Menu
```
nathan-rollins, line 12: Please rename class: Uds_Custom_Walker_Widget_Nav_Menu

inc/class-custom-walker-nav-menu.php
```diff
@@ -0,0 +1,113 @@
...
+ * Description: A custom WordPress nav walker class to customize navigation menu element wrapper and child menu's wrapper.
```
nathan-rollins, line 13: Update description: A custom Nav Walker class to generate ASU Web Standards-compliant menus for widget-based menus (sidebars, etc.)

inc/navigation-widget.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * Custom Navigation menu widget
```
nathan-rollins, line 3: Rename: Custom ASU Web Standards Navigation menu widget

inc/navigation-widget.php
```diff
@@ -0,0 +1,66 @@
...
+		$args[0]['after_widget'] = '</nav>';
+		$args[0]['before_title'] = '<div class="nav-text">';
+		$args[0]['after_title']  = '</div>';
+	}
+		return $args;
+}
+add_filter( 'dynamic_sidebar_params', 'uds_wp_theme_custom_widget_title_tag' );
+
+
+/**
```
nathan-rollins, line 28: Always wrap these functions with

```
if ( ! function_exists( 'uds_function_name' ) ) {
```

This is important to ensure child theme functionality (and ability to override if needed).

nathan-rollins (changes requested):
There are still requested changes to be addressed.

inc/class-custom-walker-nav-menu.php
zainabD: Is there a naming standards to follow? or is it a suggestion to change the class name to Uds_Custom_Walker_Widget_Nav_Menu?

nathan-rollins: It is more the case that I was trying to make the class name more explicit that this class generates ASU Web Standards nav menus.

As this [PHP wiki page](https://wiki.php.net/rfc/class-naming) shows, there is friction in the PHP community about class naming conventions. I usually follow the "Always Pascal" rule, which means the first character is capitalized and all following letters are always lowercase. So that is why I asked for "Uds_Custom..." instead of "UDS_Custom...".

I am fine with the second option if you want to use that. I just suggested the usual naming schema we have used in our other projects.

zainabD: No problem, I'll change it to Uds_Custom_Walker_Widget_Nav_Menu

zainabD:
I didn't update readme file since this is not a new feature, we're still using the **Navigation menu** core widget, with custom classes, wrappers and nav walker in order to make the side menu match the ASU Web Standards. 
Do we still want to update the readme file in this case?

nathan-rollins:
Oh, good point, Zainab. My request to the update the README was because I was also thinking about your other PR for the new sidebar widgets that didn't get a README update.

nathan-rollins (approved):

nathan-rollins:
Changes look good. PR is approved. I also renamed this PR to make its purpose more descriptive and clear. (Especially important when we browse the PR history in the future.)

PR #67, opened 2020-11-24
Zalsidiki search widget

zainabD:
Updated html tags and class for the search field on searchform.php to make it match the new standards. 

- Updated html tags and class on searchform.php
- Only used bootstrap classed
- These updates can be tested when adding a search widget. 
- I kept the widget title as it is since we still don't have enough information about the sidebar titles, as if they need to follow any standards like the the one on side menu's title.


searchform.php
```diff
@@ -10,13 +10,12 @@
 ?>
 
 <form method="get" id="searchform" action="<?php echo esc_url( home_url( '/' ) ); ?>" role="search">
-	<label class="sr-only" for="s"><?php esc_html_e( 'Search', 'uds-wordpress-theme' ); ?></label>
```
nathan-rollins, line 13: Careful, this is for screen-readers (accessibility). We need to verify whether this can be safely removed. (and if it has been removed from the example Storybook code, was that approved and confirmed to be ok?) I would ask Steve Ryan for confirmation with Kathy Marks, who has final approval on all accessibility features and markup.

nathan-rollins (commented):
We need to confirm the removal of the screen-reader label is approved. Since it is only visible to screen-readers, we probably shouldn't be touching it at all.

nathan-rollins (changes requested):
Please restore the label class="sr-only" line.

nathan-rollins (approved):

PR #72, opened 2020-11-25
Page template restructuring - page.php & landing.php

### LINKED ISSUE #61: Bug: Image asset path in compiled SASS still references old theme name
The utility classes for any of the background textures that ship with UDS Boostrap 4 aren't working out of the box.

## Steps to reproduce?

1. Install a new copy of the theme in a new environment.
1. Create a `<div class="container-fluid bg network-white" style="height:20rem;"></div>` someplace on a blank page.
1. The background image should NOT show up, and should reference the old theme name in the `background-image` attribute for the div.

The cause for this error was a change in the variable that controls the default location of those image assets without also running a `gulp watch` or other SASS compiling method once the variable had been changed. Recompiling the SASS with the variable's default value resolved the problem.

### LINKED ISSUE #64: BUG: Registered page templates in a child theme are not accessible to native UI
## Steps to reproduce?
When using a child theme or a plugin with the **UDS-WordPress** theme, you would typically have the ability to select the created page template from the native UI in either the `post.php` screen or within the Quick Links section of `edit.php.`. 

The new interface for selecting a page template (full-width, fixed-width, and either sidebar arrangement) looks great and provides an excellent way for non-native users to locate the control and layout their pages accordingly. However, because the values for the page templates are hard-coded into the theme's `functions.php` and related files, the result is the inability to select your intended template. 

## Recommendations for actions to address the issue.

- [ ] Add a couple of missing hooks/filters that would allow for a developer to programmatically overwrite that functionality. 
- [ ] Utilize `get_page_templates()` function to list out all possible registered page templates and build generic icons for each one within the interface dynamically.
- [ ] Restore the dropdown in the quick links area as a much faster method to alter multiple page templates at once.

SteveRyan-ASU:
Resolves #64 
Resolves #61

This pull request makes a number of proposed changes in order to work around a couple of noted problems.

- Blocking issue #64 affects child themes that attempt to register additional page templates. Those page templates become inaccessible in the UI.
- In general, the page templates produced by our theme contain several layers of HTML "wrappers" which make it difficult to constrain the produced content to 1920px at a maximum.
- Those same wrappers and intrinsic Bootstrap containers and rows will eventually also make it difficult to define and build a block pattern that works across all circumstances.

To attempt to provide a solution, I took a hard look at the structure that was being produced by the default `page.php` template. I also examined the pattern found within the native templates that produced a page with a sidebar. This PR focuses on those experiences as initial proofs of concept. Should we move forward with these proposed ideas, we will need to translate the learning here to the other parts of the theme. (`index.php`, `archive.php`, `author.php` and `single.php`)

## Global changes, high-level decisions
The original theme that we forked for this project included a number of global options for the default position of a sidebar element across all templates. Those global options provide a degree of customization to the default theme without the need to encode a solution. 

In our case, we have the opportunity to decide to enforce some of those defaults within our page template designs rather than rely on an end-user to configure the product correctly. 

Hence the following "let's blow it up" strategy. 💣 

- I removed all traces of a global option for a default sidebar placement. The basic option to have a sidebar or not and where to place that sidebar were reworked into other parts of the theme.
- I also reworked the number of default widget areas the theme ships with. There is no real need for a "left sidebar" area that is different from a "right sidebar" area. So, now there is only one.
- I also reworked the number of page templates that the user has available "out of the box" down to only two - one that produces a sidebar (`page.php`) and one that does not (`landing.php`).
- I also dropped the support of a "featured image" for the `page` post type, in favor of the use of our hero image settings or eventual block.

## hero.php

- I included an option in this template to display the page's title at the top of the page template if a hero image was not set. 

## landing.php

- This is a new page template that produces a page with no included Bootstrap container.
- It also does not natively produce a sidebar. 
- The intended purpose of this page template is to provide a canvas for building a block library that will feature blocks that include containers, rows, and "wrapping" elements at a high level. 

## page.php 

- The default template for a `page` within WordPress will now produce a sidebar.
- This page template includes a preconstructed container and row setup from Bootstrap in order to provide the intended sidebar-content layout. 
- It is intended to be compatible with all of the native blocks for WordPress in a manner that would not require any additional "container/row management" style elements in the markup of a page. That should make it a very straightforward editing experience for end users.
- This also enforces an opinionated manner of marking up a content-sidebar oriented page which is to place the `#primary` content above the `#secondary` sidebar content. On a mobile device, this will result in the sidebar content ALWAYS appearing below the main content.
- The mechanism for altering the orientation of the sidebar and content columns was simplified to include a single ACF field and a CSS class in the container `<main>` element on the page. (The layout can be adjusted by altering the `order` CSS property.


SteveRyan-ASU:
Formally requested a review of this PR after looking over it myself for a little while and fixing a few minor mistakes on my part.

SteveRyan-ASU:
This has the side effect of also closing #61, as it adds a SASS rule or two to the theme.

wmcconnell:
> This has the side effect of also closing #61, as it adds a SASS rule or two to the theme.

That's what I like to see. Double win!

zainabD:
Good job Steve! I was only wondering why can't we keep the metabox I built with only two options (right sidebar, left sidebar), and keep the option of selecting **which** sidebar to be displayed in which place. 
I see that you only added the option to select "where to display the sidebar" but not "which sidebar to display"

SteveRyan-ASU:
@zainabD: Elsewhere in the PR, I dropped support for one of the two default widget areas. Right now, if you look at `widgets.php` you'll only see one widget area defined. 

There were a couple of driving factors in that decision, which I didn't do a great job of outlining above.
- One reason was for simplicity. If we can switch the orientation of a sidebar via a CSS rule, there's no need to have a separately defined area for a "left" and "right" widget area. 
- With the default features in place, If you happened to want the same widget to appear within the right sidebar in one page and the left sidebar on a different page, there was a problem. The only way to accomplish that would be to create duplicate widgets and place them in the context of both widget areas. By aligning to one widget area + providing a way to control the position, you can alleviate that potential problem. (It's an edge case, admittedly.)
- With the drop of full-site editing features just around the corner (core 5.7 now?) the whole problem of widget management becomes really different looking. By simplifying this now, I think we'll be in a better place to deal with what is coming in the future. 

That does open up a different discussion about how to deal with problems like:

- How do I control which widget is visible on which page?
- And what if I just wanted to make a completely separate sidebar on this specifc set of pages?

But those problems luckily have solutions in the plugin realm that can help fix that up.

- [Widget visibility](https://jetpack.com/support/widget-visibility/) was a plugin that privided a tried and true method to set conditional rules on individual widgets in a variety of ways.  It currently ships as an option within Jetpack. 
- But Jetpack is bloated, so I abstracted that feature and 2-3 other smaller features into a utility plugin that we use across our sites without issue. See [Jetpack UnPlugged](https://github.com/asuengineering/jetpack-unplugged). 

Adding a completely new widget area "on the fly" used to be accomplished via a different plugin by WPMU called [Custom Sidebars](https://wordpress.org/plugins/custom-sidebars/) However, that plugin's lifespan is officially measured by how long it takes for WP core to be updated with the full-site editing features.

- See this announcement for the writing on the wall there: https://premium.wpmudev.org/blog/retiring-custom-sidebars/

wmcconnell (approved):
We've walked through the changes during our last meeting, and my test pages with both templates are not showing any obvious showstoppers. Approving, with new notes added to the project so we don't forget to address two things we talked about:

- Where the global banner should go on a page with no hero. I'm thinking above all content, but I'll get confirmation
- How to re-implement specific sidebar menus on individual pages when needed

PR #73, opened 2020-11-30
Fix Main Menu Warnings About Missing Array Keys

### LINKED ISSUE #70: Menu Errors With Anything Below Grandchild Level
Creating a standard WordPress menu with items below the grandchild (third) level of nesting results in multiple 'Undefined index' errors on the menu if you have debug turned on. In cases where debug is off, there are no errors, obviously, but menu items below the grandchild level do go missing.

There are a couple of things to keep in mind here:

1) I don't think the standard, as it exists now, even allows for items below the grandchild level; I don't think the code we originally wrote even looks for them. The missing items are somewhat expected behavior, then.

2) Nothing stops a user, however, from using deeper nesting in the default WordPress menu builder.

3) The solution to this may be code, but might also touch on user permissions and how much we expect even our most hands-on clients to be able to modify main navigation menus.

wmcconnell:
Resolves #70 

This pull request fixes an issue where our custom menu code was throwing warnings about missing array keys when any menu item was four, or more, levels deep in the WordPress visual menu builder (and you had debug turned on). There was only one line of code actually causing the warnings, and I applied the simplest fix possible: an `if` statement around the line to check that the array we were meaning to modify actually had the key we expected - which is true for all expected menu items, but is false once you get below three levels of nesting.

Since this issue essentially resulted from the undocumented fact  (in our theme, anyway) that we only support three levels of menu items, I also added documentation in multiple places and ways.

- New conditional code to ensure that menu items that are too deeply nested do not cause warnings, and are, in fact, not added to the final menu array during processing
- Updates to the README file with an explanation, and accompanying image, to explain how nesting your menu items in the menu builder will affect the final menu
- A new tab in the contextual help for the menu building admin page, with the same language from the README
- New admin CSS that highlights menu items more than three levels deep. This was an experimental idea, and we can remove it from this PR if the collective opinion is that it is not a good one. It works via a SCSS loop that create 10 CSS rules to highlight items more than three levels deep in the WordPress menu builder. This solution only highlights menu items at levels 4 through 14. In the case that someone ignores 14 levels of warnings, the highlighting will stop even though the menu items will not be part of the final menu.

Although this issue was discovered during testing of my menu highlighting branch, it wasn't really related to code on that branch so I'm hoping we can approve this one first (and quickly?) in order to wrap up work on that initial PR.


nathan-rollins (approved):
Nice solution!

nathan-rollins (approved):

PR #74, opened 2020-12-02
Readme Update: Special Characters and Structure

### LINKED ISSUE #49: Readme: Move 'Updating the Theme' Beneath 'Using the Theme' Section
We should move the instructions for updating the theme out of the `Installation` section, and into the `Using the Theme` section. While we do cover using the Github Updater under `Installation`, updating the theme is a regular task that someone using the theme would do.

### LINKED ISSUE #50: Readme: Remove Special Characters from Titles
We should remove the special chevron-like characters that remain in a couple of titles. I've already removed all others, and these two were overlooked.

wmcconnell:
resolves #49
resolves #50

A couple of small tweaks to the readme that I noticed too late during my initial reorganization:

* Removed remaining instances of the a special caret character in some titles, as they just made more work

* Moved the 'Updating the Theme' section (which was empty, just a header) underneath 'Using the Theme', as it made more sense to me


wmcconnell:
This is literally moving two lines of the README around. I'll wait a bit for a review, but I'm planning on merging it sooner, rather than later, so I can take a look at Zainab's work with all my stuff out of the way

nathan-rollins (approved):

PR #78, opened 2020-12-23
Fix Default Values and Modernize the Customizer

### LINKED ISSUE #61: Bug: Image asset path in compiled SASS still references old theme name

### LINKED ISSUE #62: Footer Menu Showing Errors on New Installs
New sites in our conversion multi-site show errors in the footer right after the site is created. The errors indicate that we are attempting to loop through menu items that don't seem to exist. This can be remedied by manually associating the menu with the footer menu area in the WordPress admin.

At our most recent meeting, we discussed the probability that this was an artifact of recent changes to how the default menu items are created.

### LINKED ISSUE #77: Provide support for a link in the main title of the site.
## Issue is about..
  - [ ] a Bug Report
  - [x] a Feature Request

## Problem statement
The UDS documentation for the header says that:

From: https://unity.web.asu.edu/@asu-design-system/bootstrap4-theme/index.html?path=/docs/docs-global-header-header-main--page

> There are two supported configurations of [title information in the header]. Both the parent titles and the main site titles can be either linked or unlinked in either configuration.

Currently, our theme doesn't provide support for a link to be placed within the markup for the header in this way. 

# Todos

- Add a checkbox to the customizer settings for the theme that would allow the end-user to toggle on/off this link.
- The link should always default to the home page of the site - no need to solicit more user input.

wmcconnell:
Resolves #61
Resolves #62
Resolves #77 

While working on resolving the default settings issues listed above, I noticed that part of the problem was that we were using both 'theme_mods' _and_ normal options values to hold our theme settings. While doing some research, I saw this in the theme documentation (emphasis mine):

> There are two primary types of settings: options and theme modifications. Options are stored directly in the wp_options table of the WordPress database and apply to the site regardless of the active theme. **Themes should rarely if ever add settings of the option type.** Theme mods, on the other hand, are specific to a particular theme. Most theme options should be theme_mods...

So I moved all of our settings into theme_mods, which puts all of them in the options table under one entry, named `theme_mods_UDS-WordPress-Theme`. The problem was that references to the old options were all over the customizer and in our templates, so I ended up needing to update the customizer and template files. In doing that, it seemed like a good idea to bring the customizer up to modern WordPress standards. 

- Change all of our theme's settings to theme modifications, and update the customizer and the affected templates to use those mods via the theme modification API in WordPress
- Move the display logic for parts of our templates controlled by Customizer settings into functions, so we can implement what WordPress calls 'selective refresh' in the customizer. See [this page from the WordPress docs](https://developer.wordpress.org/themes/customize-api/tools-for-improved-user-experience/#selective-refresh-fast-accurate-updates) for more
- As a result of the previous item, get visual, discoverable icons in the customizer to show you where you can change things. Clicking on a button brings you right to that setting, with the appropriate control selected and ready to change

https://user-images.githubusercontent.com/31013359/103044979-e8726780-453f-11eb-8ad8-33caf68b4e4c.mov

- Re-organize the customizer settings, moving some to new sections where I thought it would make sense (things in the footer are now under ASU Global Footer, etc.), removing some, and re-ordering some within their section

I'm sure there will be some changes, but we've got a good start here for an updated, modern Customizer that's easier for our clients (and ourselves) to use.


zainabD:
Good job Walt!
I have few things that I noticed in the customizer: 
1- Under **Site Information**, can you hide these two fields ["Parent Unit" and "Parent Unit URL"] in mobile view since their content will only be visible on desktop view. 
If this wasn't doable, can you add a note next to "Parent Unit" and "Parent Unit URL" that indicates that these fields will only show in Desktop view.
2- Under **ASU Global Footer**, add the edit button to site name under <h5>, the same edit button that was added to the "subdomain-name" class in the header.
3- Under **ASU Global Footer** >> first section >> "Unit Endorsed Logo URL", can this field be dependable on value selected in the selection field above? (to make it visible when "none" is selected).
4- Add edit button next to the main menu in the header that links to **ASU Global Header** section.

footer.php
```diff
@@ -12,8 +12,8 @@
 $footer_row_branding   = 'enabled';
 $footer_row_actions    = 'enabled';
 
-if ( is_array( get_option( 'uds_wp_theme_options' ) ) ) {
-	$c_options = get_option( 'uds_wp_theme_options' );
+if ( is_array( get_option( 'theme_mods_UDS-WordPress-Theme' ) ) ) {
```
nathan-rollins, line 15: Do we still need this $c_options variable and the $footer_* variables? Looks to me like you were able to complete remove/refactor the code that used to use these variables. I don't see them being used anywhere after they are initialized.

inc/render-partials.php
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/**
+ * Render 'Contact Us' link in the footer
+ *
+ * Takes the value of the 'contact_url' setting and, if it is not empty,
+ * renders a link with that value as the URL.
+ */
+function uds_wp_render_contact_link() {
+	// =============================
```
nathan-rollins, line 148: Looks like an extraneous comment line copy-pasted from the old file. Probably should be dropped.

inc/render-partials.php
```diff
@@ -0,0 +1,192 @@
...
+ * Render the info+menu links row in the footer
+ *
+ * Determines whether or not we render the info+menu row of the footer.
+ * When this is set to 'disabled' we skip generating the markup altogether,
+ * meaning that individual settings for contact, contribute, and the menu
+ * items are ignored altogether.
+ */
+function uds_wp_render_footer_action_row() {
+	$action_row_status = get_theme_mod( 'footer_row_actions' );
+	$org_name = wp_kses( get_bloginfo( 'name' ), array() );
```
nathan-rollins, line 168: Wrapping get_blogingo() in wp_kses() is unnecessary. get_bloginfo() supports filtering the property being retrieved (though 'raw' is the default filter option.)

See [get_bloginfo Function Page](https://developer.wordpress.org/reference/functions/get_bloginfo/)

nathan-rollins: But it is a good question for us to ask. Do we want/need to filter our Site Name property to prevent embedded html tags? If so, then we should set the filter option here and wherever we use get_bloginfo('name'). We have this inc/render-partials.php (line 23).

And we should probably reuse that render function here anyway.

nathan-rollins (changes requested):
Have a look at my questions/notes.  Everything looks pretty good fro here, just a couple possible tweaks.

js/customizer.js
```diff
@@ -8,35 +8,55 @@
...
-				$( '.site-title a, .site-description' ).css( {
-					'color': to
-				} );
+				if ( 'disabled' === to ) {
+					$( '#menubar' ).empty();
+				} else {
+					$( '#menubar' ).html( '<span class="nav-link" href="#" title="Example Link">Menu will appear here after you publish and exit the Customizer</span>' );
+					$( '#menubar' ).css( 'display', 'flex' );
+				}
 			}
```
zainabD, line 58: Can you try to apply the same function as applied on the footer to have the nav menu called instead of displaying the line "Menu will appear here after you publish and exit the Customizer".

inc/customizer/customizer-settings.php
```diff
@@ -37,475 +45,520 @@ public function render_content() {                  ?>
...
-		// = Remove Default WordPress Customizer Controls and Sections  =
-		// ==============================================================
-		// ==============================================================
+		// Remove default sections and controls we do not need/want.
 		$wp_customize->remove_section( 'background_image' );
 		$wp_customize->remove_section( 'colors' );
 		$wp_customize->remove_section( 'header_image' );
 
+		$wp_customize->remove_control( 'blogdescription' );
 		$wp_customize->remove_control( 'site_icon' );
```
zainabD, line 54: Do we also need to remove blogdescription field from **Settings >> General** section?

zainabD (changes requested):

inc/customizer/customizer-settings.php
nathan-rollins: Good question. It was removed because there was no use for it in the new Web Standards (it used to be displayed in the old ASU footer.) 

But if you guys have a good use case for it, go ahead and bring it back it. If so, we need to document where the Blog Description property is to be displayed in the site framework. It is not actually rendered anywhere right now.

wmcconnell: I think Zainab is asking if we should add code to _also_ remove this field from Dashboard->Settings->General. While we have removed this setting from the Customizer, it can still be set via general settings. I don't know if it's expected behavior for a theme to remove standard admin controls like that, but since we're not actually using this value, I can see how it might confuse someone if they set the tagline and it has no effect.

nathan-rollins: Ah, brilliant!

zainabD:
> 3 - Thank you for pointing this out. At this time, this property isn't the "URL link" for the endorsed logo. It is the old Customizer option from our old theme framework--applying an alternate logo by pasting a URL to an external image file. We kept this "Alt Logo" property as a backup to the endorsed logo Dropdown option. See "/inc/render-partials.php - uds_wp_render_footer_logo()" for the current footer logo logic.
> 
> But this property in the Customizer is pretty confusing isn't it? Especially, since we now need a Link Url property for the footer logo.
> 
> Do we want to remove this old Logo Url property (or change it into the Link Url we now need?)
> 
> If we do this, then our theme can only use endorsed logos that have been officially added to the framework. We won't have a quick way to use an alternate logo not in the prebuilt dropdown. That isn't a big deal for us or Engineering, I would guess, but it also means other ASU WP teams will not be able to use their unit logos until they are added to our theme. (Unless you can come up with a better way to quick add an alternate logo?)

Thank you Nathan for the clarification, but from my understanding, this alternate logo field will not be used unless the "none" value was selected from the selection list above, that's why I thought it would be more reasonable if we make this field a dependable field if possible. What do you think?

nathan-rollins:
I agree, this needs to be improved. Your idea sounds good to me.

zainabD (dismissed):

nathan-rollins (approved):
Cool solution!

wmcconnell:
Pushed one - final - commit for this PR. I think some of my local dev issues may have resulted from differences between the new customizer settings I had created (or renamed), and the defaults we were setting. I went back and set defaults for the new/renamed customizer settings. Seems to be working. To test this, I would switch themes, delete all of our theme-related options and theme_mods, then activate our theme again. Tried this three times without any trouble.

zainabD (dismissed):

SteveRyan-ASU (dismissed):
I've made a few changes here to restore some lost functionality. I obviously approve of this PR now, but it should probably have another set of eyeballs to check.

SteveRyan-ASU:
After reviewing the code from an end-user perspective on a new install of WordPress, I discovered that a couple of the new functions that were being used in the customizer were producing markup that was inconsistent with that which was recommended in UDS-Boostrap for the header.

No fault of anyone except maybe mine. =) That header markup is a bear, and now I can also point out some inconsistencies in the way that I encoded some options that could be ironed out in the future. 

Here's what I addressed:
- When showing/hiding the navbar part of the header, there needed to be a CSS class added or dropped in `#navbar-container` along with the decision to render or not render the actual bar.
- When showing/hiding the site title or parent site information, there was a class and some markup that needed to change depending on the scenario. Those variations were also included in the fix.

SteveRyan-ASU (approved):
I am Steve Ryan and I approve of this pull request. ™️ 😁

wmcconnell:
Okay, finally putting this one to bed! Thanks to everyone who pitched in with code and reviews!

PR #79, opened 2020-12-30
Deleted bg-transparent class from the search button

zainabD:
I noticed that the bootstrap class "bg-transparent" was added to the search button which gives it a transparent background. 
I only removed the class from there so that when we add a background to the 404 page, the search bar will be displayed all as one bar including the search button, instead of having the search button looks separate when the transparent background is applied.


wmcconnell (approved):
Looks good. We've already gone over this in one of our working sessions, so I've seen it work. Approved!

PR #84, opened 2021-01-19
Added a checkbox to have the option to hide page title

zainabD:
I noticed that we need to have the option to hide page title in case if page title has to be styled like the ones in Devils drop off website.
I had to remove the page title and add a regular Heading block to create a styled title. But the issue was that the pages would show with no title in the editor, and we can't tell which is which: 
 
Screen Shot 2021-01-19 at 1 57 48 PM

To solve this issue:
- I created a True/False field in ACF.
- Added an if condition before printing the page title in **hero.php** template and **content-single.php**. 
- This feature will be added to post and page types. 

Screen Shot 2021-01-19 at 2 03 04 PM

### DIFF
--- a/acf-json/group_60074496e95ba.json
+++ b/acf-json/group_60074496e95ba.json
@@ -1,0 +1,50 @@
+{
+    "key": "group_60074496e95ba",
+    "title": "Page title",
+    "fields": [
+        {
+            "key": "field_6007449f54b72",
+            "label": "Hide page title",
+            "name": "hide_page_title",
+            "type": "true_false",
+            "instructions": "",
+            "required": 0,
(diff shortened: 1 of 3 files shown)


wmcconnell (approved):
Nice, clean solution. Works as advertised, and we need this quickly for the Devils' drop-off site. Approved with the comment that we should probably set some time aside to work out a consistent visual language with our ACF fields so that our various controls and blocks are uniform across the theme.

wmcconnell:
As an example of the visual style question, I have been using a style more like this in my ACF Pro blocks:

Screen Shot 2021-01-19 at 3 06 15 PM

PR #87, opened 2021-01-20
ACF Block Setup and Samples

wmcconnell:
Since we'll be moving forward with ACF Pro for our custom blocks (at least at the time I'm writing this), I thought it would be a good idea to come up with a way to organize the related files. I've created some sample ACF Pro blocks to demonstrate a suggested organization, as well as document the basic methods for creating blocks with ACF Pro for those of us who have not used them yet.

The basic approach here is to keep our blocks in their own, individual, folders inside a new `templates-blocks` folder that lives alongside the other template folders. Each block will have its own `register.php` file that does the work of registering the block type(s) ([see the ACF docs here](https://www.advancedcustomfields.com/resources/acf_register_block_type/)), and a template file named after the block itself, like this:

```
\templates-blocks
     \my-block
          register.php
          my-block.php
```

In the case of related blocks (say multiple types of cards), we could register them all in one `register.php` and keep the together in the folder, like this:

```
\templates-blocks
     \uds-cards
          register.php
          standard-card.php
          event-card.php
          ...
```

- add a new file, `uds-blocks.php`, to load/register our ACF blocks, and call it from `functions.php`
- add a new folder, `templates-blocks` where all of our ACF block templates will live, inside their own individual 
- add two sample blocks, one that uses the 'traditional' ACF way, and one that allows for more customization through the use of the `<InnerBlocks />` tag
- add ACF Pro JSON files for the sample block fields

_Note: The sample blocks are here just to show the organizational structure, basic ACF block coding techniques, and potential approaches to building blocks with ACF. They will be deleted once we have added some actual blocks. Feel free to comment on them, but this PR is about setting up the organization of the files, and not these specific blocks._


zainabD (commented):
This is a good way to organize our blocks and have them all stored in one place, and called from one file.
I know that these are only sample blocks and will be deleted once we have a real block created, but there is one thing I noticed, is that when I try to select the **sample inner block** from the gutenberg block list, the list will collapse immediately, it will not let me select the block. 
I'm not sure if this is an issue with ACF or if there is something wrong with that particular block since it didn't do it with the other sample block.

nathan-rollins (dismissed):
Makes perfect sense to me. Looks great!

wmcconnell:
Zainab,

That's weird. I'm able to select that block. I wonder if it's related to the field groups? When we have new blocks, you may need to go to your ACF plugin and sync some new JSON to your local database.

https://user-images.githubusercontent.com/31013359/105425513-c59ea600-5c06-11eb-9377-144b8d2c2605.mov

SteveRyan-ASU:
This looks good from my perspective as well. I appreciate the guidance about how & where to register blocks within the theme. It'll keep the codebase way more organized.

I did notice a flicker when I hovered over the same icon that @zainabD noted. Screen capture attached. Note that I never actually got to click on the sample block icon.

SteveRyan-ASU (approved):
Approved with the following intent:

- The TGM plugin activation script should now reflect a hard requirement for ACF Pro upon theme activation now that we are actually shipping ACF blocks with the theme.
- I'll raise an additional issue to add that requirement.

zainabD:
> This looks good from my perspective as well. I appreciate the guidance about how & where to register blocks within the theme. It'll keep the codebase way more organized.
> 
> I did notice a flicker when I hovered over the same icon that @zainabD noted. Screen capture attached. Note that I never actually got to click on the sample block icon.
> 
> 

We figured that the issue occurs on Chrome only, what browser were you using @SteveRyan-ASU ?

wmcconnell:
I have pushed some changes that should fix that issue. Long story short, I was trying to preview a block that has no content (it's just a container for your inner blocks). I set the 'mode' of that little preview window to 'edit', to show the actual ACF controls, instead of the missing content, and it seemed happier.

zainabD (approved):
That fixed the issue, nice work Walt!

Issue #81, opened 2021-01-19
Format native blocks - Block Category: Text


SteveRyan-ASU:
Stemming from a block planning conversation between the KE team and the Engineering team, we came up with the following punch list of stuff to do regarding the native blocks that ship with WordPress. The majority of this work involves some testing within the default template (sidebar) to make sure that they work as expected out of the box. 

Here's what we determined:

### Block Category: Text
Keep: paragraph, list, code, table, classic block, header
Remove: preformatted, verse, pullquote

#### Pull quote vs blockquote
We only need one of these two blocks for right now. Recommend that we turn off the **pull quote** block for now as the formatting for the pull quote is essentially a `blockquote` element wrapped with a specific wrapper for WordPress.

The blockquote element within UDS has several layers of color combinations possible. The recommended classes for these color combinations within the UDS-Bootstrap project are... not great... so we may be better off reinventing the wheel a little bit within our theme if that's easier. 

#### Header highlight classes
Cross-reference this list with a separate issue opened to add the `<span class="highlight-gold">` classes to the middle of a header. That would be issue #53.

SteveRyan-ASU:
Quoting a discussion from PR #88, we perhaps should consider the following actions when thinking about registering/unregistering blocks. (I had no idea that this was possible via JS initially.)
 
> ### If we think we will get into a situation in which we would want to restrict blocks to a specific page template, then the JS required gets immediately more complicated.
> * See this article for the full description of why: https://webservices.ufhealth.org/2020/07/08/how-to-restrict-blocks-to-specific-page-templates-in-the-wordpress-gutenberg-editor/
> * TL;DR: Since the page template can be switched "on the fly" from within the editor, there are LOTS of things we should potentially do to check the state of the allowed/not-allowed blocks upon the user clicking the dropdown.
> 
> ### The process for unregistering blocks from the editor via PHP appears a lot easier. However, there is a catch.
> * Here's the filter: https://developer.wordpress.org/reference/hooks/allowed_block_types/
> * The second parameter of that filter also allows you to reference the allowed block types list against a specific `wp_post` object.
> * The `wp_post` object contains references to individual pages, page slugs or specific post types. But not to the page template within the same post type.
> 
> ### What I'm wondering is the following:
> * Should we consider moving the `landing.php` template out of the native post type `pages`.
> * ... and instead, create a custom post type called `Landing` with the same intent & markup as the landing.php template.
[...]

PR #88, opened 2021-01-21
Clean up files that were created previously for the sidebar metabox that was overwritten by Steve's work in #72

zainabD:
I added a new Javascript file called **admin-blocks.js**, enqueued it in admin side, so that we can place all block related functions in there (functions like register block, unregister block, render block, enqueue assets and etc

- A new .js file called **admin-blocks.js**, where I unregistered 3 blocks per request [#81](https://github.com/asu-ke-web-services/UDS-WordPress-Theme/issues/81) 
- Updated **inc/enqueue.php**, called the new js file in admin side using admin_enqueue_scripts action, and deleted the call of metaboxes.js, since these functions were used for the sidebar metabox that was overwritten by Steve's work in [#72](https://github.com/asu-ke-web-services/UDS-WordPress-Theme/pull/72).
- Deleted the Metaboxes.js for same reason that was mentioned above.


inc/enqueue.php
```diff
@@ -67,11 +67,11 @@ function uds_wp_metaboxes_scripts() {
 		global $pagenow;
 		$the_theme     = wp_get_theme();
 		$theme_version = $the_theme->get( 'Version' );
-		$js_metaboxes_version = $theme_version . '.' . filemtime( get_template_directory() . '/js/metaboxes.js' );
+		$js_admin_blocks_version = $theme_version . '.' . filemtime( get_template_directory() . '/js/admin-blocks.js' );
 		$css_metaboxes_version = $theme_version . '.' . filemtime( get_template_directory() . '/css/admin.min.css' );
 		if ( 'post.php' == $pagenow || 'post-new.php' == $pagenow ) {
 
-			wp_enqueue_script( 'uds-wordpress-sidebar-metabox-scripts', get_template_directory_uri() . '/js/metaboxes.js', array( 'jquery' ), $js_metaboxes_version, true );
+			wp_enqueue_script( 'uds-wordpress-sidebar-metabox-scripts', get_template_directory_uri() . '/js/admin-blocks.js', array( 'jquery' ), $js_admin_blocks_version, true );
```
wmcconnell, line 74: It looks like we left the original file handle of 'uds-wordpress-sidebar-metabox-scripts' in there. While it would still work (as long as that name is unique), we should probably rename this to avoid confusion since this is not a script about meta boxes. I would suggest something like 'uds-wordpress-admin-blocks' to replace that here, and in line 75 below.

inc/enqueue.php
```diff
@@ -67,11 +67,11 @@ function uds_wp_metaboxes_scripts() {
...
 			wp_enqueue_style( 'uds-wordpress-sidebar-metabox-styles', get_template_directory_uri() . '/css/admin.min.css', array(), $css_metaboxes_version );
```
wmcconnell, line 75: Just a reminder to use the same script handle from line 74 above, once we have changed it to better match what this script actually does.

wmcconnell (changes requested):
Could we rename the handle for this new script, per my comments? Since this is not actually a script that deals with meta box styles, we should use a handle that matches what the script is doing - just to avoid confusion.

inc/enqueue.php
zainabD: This css file only has the metabox related styles, so I'm not sure if we need to rename it or to just delete the file since it's not being used at the time.
What do you think?

zainabD (commented):
I updated the function name and the new script name but wasn't sure about the css file since it only has the metaboxes related styles.

SteveRyan-ASU:
Man, did this open up a can of worms on my end... totally learned some stuff about registering and unregistering blocks that I didn't know before. 🤷‍♂️ 

Pertaining to the work done against the requirement from #81 to remove blocks, here's what I am wondering about:

### If we think we will get into a situation in which we would want to restrict blocks to a specific page template, then the JS required gets immediately more complicated.

- See this article for the full description of why: https://webservices.ufhealth.org/2020/07/08/how-to-restrict-blocks-to-specific-page-templates-in-the-wordpress-gutenberg-editor/
- TL;DR: Since the page template can be switched "on the fly" from within the editor, there are LOTS of things we should potentially do to check the state of the allowed/not-allowed blocks upon the user clicking the dropdown.

### The process for unregistering blocks from the editor via PHP appears a lot easier. However, there is a catch.

- Here's the filter: https://developer.wordpress.org/reference/hooks/allowed_block_types/
- The second parameter of that filter also allows you to reference the allowed block types list against a specific `wp_post` object.
- The `wp_post` object contains references to individual pages, page slugs or specific post types. But not to the page template within the same post type.

### What I'm wondering is the following:

- Should we consider moving the `landing.php` template out of the native post type `pages`.
- ... and instead, create a custom post type called `Landing` with the same intent & markup as the landing.php template.
- ... so that this whole process of limiting blocks to specific regions of WordPress becomes easier.

zainabD:
Per our last discussion with Steve and Walter, we decided to handle the block registration functions using php instead of JS, therefore we don't need the js file that was created in this branch, we only need to remove all the remainings of the sidebar metabox from old branch (this was overwritten by Steve's work in #72).
- I deleted all files, functions and classes that were created previously for the sidebar metabox that was overwritten by Steve's work in #72.
- Deleted **js/metaboxes.js** 
- Deleted function '**uds_wp_metaboxes_scripts**' from **inc/enqueue.php** 
- Deleted css class that were created for the sidebar metabox from **css/admin.css**

wmcconnell (approved):
Looks good. Approved!

PR #112, opened 2021-01-28
Create "landing" custom post type. Explore what happens next.

SteveRyan-ASU:
Work in progress. 

- Custom post type created and worked into the `functions.php` stack at a good place.
- Remembered that custom post types all come with an added slug in their URL. So all pages created within this CPT would now become `sitename.asu.edu/landing/page-title` which seemed unacceptable.
- Straight up rewriting the URL to `/` via the `register_post_type` function apparently no longer works. 
- Stumbled into here: https://wordpress.stackexchange.com/questions/203951/remove-slug-from-custom-post-type-post-urls
- Hacked up a working solution that indeed removes the slug from the URL and does not cause a 404 error.

Then: I realized that the slug rewrite code also prevents the correct page template from loading in the hierarchy. The code that's currently commented out will rewrite the slug, but will also cause the page to be loaded from `page.php` instead of `single-landing.php` like it should be.


SteveRyan-ASU:
Moving this draft PR in a new direction due to the difficulty of creating a custom post type with a default slug of _null_. 

Instead, I've created a CPT called `document` which should translate into our sidebar oriented template. Once that was created and placed into the functions.php stack, I also reworked the two page templates included with the theme to fit within the CPT. So what was formerly at `page-templates/landing.php` is now `page.php` and the new custom post type template that includes the sidebar is at `single-document.php`. I also made a small adjustment to the ACF settings to include the correct metabox information for the sidebar in the correct place.

SteveRyan-ASU:
Cross post from Teams: 

> 
> A summary of where we are at in terms of our options to implement page templates within UDS-WordPress. I apologize that this is not more straightforward. Here's where we have arrived... options in bold, potential next steps and obstacles in bullets below. 
>  
> Option 1: Use page templates the way that they are right now.
> 
> - NEXT: Find a way to resize the width of the editing window when the page template changes.
> - NEXT: Remember that the way that we'll be preventing mistakes in the future is through good documentation. Plan to write good documentation.  
> 
> Option 2: Move the sidebar-oriented template to a CPT (called a document possibly) and leave the landing page template as the default "page" experience. 
> 
> - PAIN POINT: The resulting CPT would need to have a slug (possibly docs) within the URL. No getting around the fact that a slug needs to be there. But the slug could be anything and might also possibly be able to be user-defined.
> - 95% completed with this concept right now within my PR draft.
> - NEXT: Prepare the whitelist of allowed blocks per CPT and park that within the theme.
> 
> Option 3: Ditch the page templates idea completely. 
> 
> - The concept of a page template within Gutenberg isn't very well supported in the Gutes right now. Knowing that, we may do well just to avoid that feature completely.
> - NEXT: Make page.php into the landing page template.
> - NEXT: Deliver the idea of a sidebar-oriented page as a block pattern to be deployed by an end user. This would probably result in a couple of different patterns: sidebar-left, sidebar-right, etc.
> - NEXT: Focus on finding or building Gutenberg blocks that would replicate a sidebar menu. Options there include a "deliver a whole widget area" block, or a "deliver a native menu" block.

SteveRyan-ASU:
Option 3 is starting to feel like the best way forward. Going to encode some of the basics in a new PR with that in mind.

Rationale:

- Page templates are not well supported and will probably be even less well supported once full-site editing is a thing. (Full site editing is slated for WP 5.7.
- Custom post types are ultimately very useful when your content is indeed different than a normal post/page.
- And if all sidebar pages were, in fact, used for documentation, then option 2 might make some sense. But not now.

PR #114, opened 2021-01-29
Enqueued theme.css to load in editor area only

zainabD:
I added uds_wp_gutenberg_css function to enqueue **theme.css** in editor area, this will load theme styles in admin side to show the real style of gutenberg blocks in the editor area.


SteveRyan-ASU (approved):

SteveRyan-ASU:
This looks good to me. :-) 

I grabbed some unimplemented code from the UDS-Bootstrap and tossed it into a custom HTML block. Previewing the block produced the intended result from directly within the Gutenberg editor. It also now respects layout choices that are implemented by custom HTML blocks that contain container/row/column markup.

wmcconnell (approved):
Looks good and worked for me. I imagine we may have to tweak things as we go, but let's get this merged in to help with our block development. Approved!

PR #115, opened 2021-01-30
Remove all page templates from the theme.

SteveRyan-ASU:
Remove page templates completely from the theme in anticipation of driving all page layout concerns via Gutenberg and block patterns. Of immediate concern is creating a layout solution to deal with the implementation of a sidebar as a deployable block.

To Do:
- [x] Remove page templates and reset `page.php` back to the "landing page" template.
- [x] Unregister and delete unneeded ACF and CSS code for dealing with the sidebar position.
- [x] Create location and method to unregister native blocks from all parts of the theme.
- [x] Create a shortcode that would produce a native WordPress menu. Plan to utilize the shortcode block to produce our sidebar menu until full site editing becomes the norm.
- [x] Unregister the sidebar completely once the above bullet is complete. If we are pushing things directly to the block level we won't really need a widget area for anything. (This action will set us up better for full site editing in the future.)


SteveRyan-ASU:
Added a shortcode that successfully produces a `<nav class="sidebar accordion">` element on the page. Combined with additional blocks to constrain the width of the container, it should make for an effective widget area replacement.

The shortcode looks like this: `[uds-sidebar-menu]` and accepts two options.

- `menu` which is the name, slug or id of a WordPress menu object.
- `title` which is the optional string that sits at the top of the menu and is wrapped by the `nav-text` class.

To test it out:

- Select a Shortcode block and put it on the page.
- Optionally, use some container / row / col-4 blocks for proper sizing.
- Drop the following shortcode into the block to see the example footer menu rendered as a sidebar `[uds-sidebar-menu title="Example Sidebar" menu='Example Footer Menu']`

SteveRyan-ASU:
This should be in a reviewable state now, so yay! 😁 

A couple of additional notes on the last bullet point above.

- I removed the sidebar registration from `widgets.php` but didn't do anything to modify the code that @zainabD created in `navigation-widget.php`
- Two of the included three functions in `navigation-widget.php` were necessary to filter the menu options correctly when deploying it as a shortcode (or a widget in a widget area). 
- `navigation-widget.php` also contains the include statement for the sidebar nav walker class was also included in that template file. That walker class is still in use by both the navigation menu widget and this new shortcode.

The [first function within `navigation-widget.php`](https://github.com/asu-ke-web-services/UDS-WordPress-Theme/blob/0ec86e033a703da8a9d01249727949737c5cd00d/inc/navigation-widget.php#L13) could have been eliminated, but I chose to leave it behind. Should we decide to add back a widget area/sidebar menu experience (in a child theme, for example) the navigation widget will work correctly when deployed in that context.

wmcconnell (approved):
I've pulled this down and created some test pages using the new sidebar menu short-code. Didn't run into any issues. This work, combined with Zainab's recent editor styling, should help with moving our focus to block/pattern creation. Approved.

PR #119, opened 2021-02-10
TGMPA changes to require ACF Pro and Bootstrap Blocks

SteveRyan-ASU:
Pretty much just as it sounds.:-) 

Now that we have internally decided to rely on these two plugins as "must-haves" in order for stuff to work correctly, it's time we update the TGMPA script to reflect those choices.


SteveRyan-ASU:
Still in a draft mode, but the scripts should be working correctly against a brand new installation of WordPress. :-)

zainabD (approved):
Tried this on a new installed wp and it worked, it gave me a message to install the required plugins and showed both plugins to install and activate.
Screen Shot 2021-02-12 at 12 18 00 PM

PR #120, opened 2021-02-12
Fix WordPress Admin Bar 'Quicklinks'

wmcconnell:
An update from the UDS Bootstrap theme, which we recently integrated, was causing a problem with the WordPress admin bar. New styling for the UDS "quicklinks" - links that sit at the top of the page and allow for quickly jumping down to named anchors - was conflicting with an area of the admin toolbar. Both the UDS links, and the links in the toolbar, were using the same CSS classname of `quicklinks`.

The result was that the quicklinks in the admin bar were ending up on top of the global ASU header, causing them to both be nearly invisible **and** to prevent standard WordPress admin menus from working.

Steve Ryan plans on suggesting that the UDS theme change their class name to `uds-quicklinks`, or something similar. In the meantime, I've applied a single CSS rule to the admin bar to fix this.

- Apply a `display: block;` rule to the affected ID to fix this until things are renamed upstream


SteveRyan-ASU (approved):
Good fix, well documented and easy to find in the code. 

I'm going to make the PR for the corresponding fix in UDS-Bootstrap today to change `.quicklinks` to `.uds-quicklinks`. But there's some additional worry about how that change is documented out the rest of the community.

zainabD (approved):
Worked with no issues.

PR #121, opened 2021-02-12
Bug fix: .dropdown-pinned class being applied to the wrong viewport

SteveRyan-ASU:
The global header component from UDS-Bootstrap produces a class called `.dropdown-pinned` for the mobile menu. That class is currently being applied to the menu as a whole even when the viewport is displaying the desktop version of the header.

That class applies `position: fixed;` to the whole page which resets the flexbox layout of the remainder of the page. 

This patch applies an override to that CSS rule wrapped in the correct media query so that the position property is only changed when the mobile menu is being displayed.


wmcconnell:
We ended up creating a new PR for this fix after having some issues resolving conflicts. Closing this one and approving the other.

PR #122, opened 2021-02-12
Bug fix for dropdown-pinned included.

SteveRyan-ASU:
The global header component from UDS-Bootstrap produces a class called .dropdown-pinned for the mobile menu. That class is currently being applied to the menu as a whole even when the viewport is displaying the desktop version of the header.

That class applies position: fixed; to the whole page which resets the flexbox layout of the remainder of the page.

This patch applies an override to that CSS rule wrapped in the correct media query so that the position property is only changed when the mobile menu is being displayed.


wmcconnell (approved):
Approving this one, since we went over in during a working session and the original PR was already approved by Zainab.

PR #125, opened 2021-02-16
Add White Text and Highlighting Options

wmcconnell:
To accommodate the RTO hero design, added the ability to have white text over a hero image.

- New ACF Pro field for choosing black or white text. This field only shows when a highlight style of 'none' has been chosen.
- Add a new 'white' highlight option for the hero title. This is described in the XD file, but was overlooked during the initial creation of the hero
- Updated the Hero template to accommodate these new field settings.


templates-global/hero.php
```diff
@@ -22,17 +23,43 @@
...
+ * The text color of the hero title is usually black. However, when 'none' is
+ * chosen, we allow the user to choose between black or white text to best fit
+ * with the image.
+ */
+if ( 'white' == $title_color && 'none' == $hero_highlight || empty( $hero_highlight ) ) {
+	$title_color_class = 'text-white';
+} else {
+	$title_color_class = '';
+}
+
```
zainabD, line 36: 'black' option is missing here but was added to the ACF json file. 
`elseif ( 'black' == $title_color && 'none' == $hero_highlight || empty( $hero_highlight ) )`

Also, I would remove the other side of the condition `('none' == $hero_highlight || empty( $hero_highlight ))`, since the text color option won't show unless the $hero_highlight is none anyway.

templates-global/hero.php
```diff
@@ -22,17 +23,43 @@
...
-	<div class="container uds-hero-container">
+<div class="uds-hero <?php echo $hero_size_class; ?>" style="background-image: linear-gradient(180deg, #19191900 0%, #191919c9 100%), url('<?php echo wp_kses( $hero_asset_url, wp_kses_allowed_html( 'strip' ) ); ?>')">
+	<div class="container uds-hero-container lazyloaded">
 		<?php
 		if ( ! empty( $hero_title ) ) :
 			?>
 		<h1 class="col-md-8">
-			<span class="<?php echo $hero_highlight; ?>"><?php echo wp_kses( $hero_title, wp_kses_allowed_html( 'strip' ) ); ?></span>
+			<span class="<?php echo $hero_highlight; ?> <?php echo $title_color_class; ?>"><?php echo wp_kses( $hero_title, wp_kses_allowed_html( 'strip' ) ); ?></span>
 		</h1>
```
zainabD, line 63: Since it's either **$hero_highlight** or **$title_color_class**.
We can create one var called **$hero_title_class**, by default it takes the value of **$hero_highlight**, and would take the value of the **$title_color_class** if it went through the if condition `if ( 'white' == $title_color) or black`.

zainabD (changes requested):

wmcconnell:
I've updated the logic based on Zainab's feedback, and cleaned up the fields a bit to remove stuff I was doing to 'improve' the hero (I'll add that all back in a proper, dedicated pull request at a later date). I renamed some fields in the ACF group, but I think those are just labels and shouldn't affect the actual performance.

zainabD (approved):

PR #127, opened 2021-02-17
Fix Maximum Width for Content Area

wmcconnell:
A typo had set the maximum width of our content area to 1960px, instead of the specified 1920px. Updated `_theme.scss` with the correct value. Changes include:

- Underneath `main > *`, changed from `max-width: 1960px` to `max-width: 1920px`
- Alphabetized some style declarations (it was my fault they were not already that way)
- Added spaces in all declarations, so that we have `margin-right: auto`, and not `margin-right:auto`, for example. I think it's easier to read


earlalex:
Reviewed in our working session.

earlalex (approved):
Looks good.

PR #129, opened 2021-02-20
Add FontAwesome Icons in Editor

wmcconnell:
Updating the admin CSS to include the same FontAwesome SVG rendering that exists on the front end. It was just a matter of enqueuing the same Javascript. Now we can see icons when placing them in cards, buttons, or wherever else we might use them.

- Enqueue FontAwesome JS in the editor
- Add a specific CSS rule to icons in the editor to match the vertical alignment adjustment that occurs on the front end


SteveRyan-ASU (approved):
Looks great @wmcconnell. Glad that we got this working, and am glad we're still using the JS way of doing things even in the editor. More parity between the editor and the front-end = a better experience. 👍

PR #130, opened 2021-02-22
Theme cleanup activities

### LINKED ISSUE #58: Complete the sidebars feature by removing any templates or deprecated functions
Now that Zainab's new Page Sidebars widgets are implemented, we need to review which page templates are still needed. I supsect we can probably get rid of most of our page templates: Left Sidebar, Right Sidebar, and so on. We just need the default template, and *maybe* a full-width template.

Additionally, any of the Customizer logic from UnderStrap for setting the default theme settings, like sidebar positions are now redundant, and we need to remove all related code and functions from the theme.

Let's do this as soon as possible so we don't forget and leave the confusing redundant code in there.

## Issue is about..
  - [ ] a Bug Report
  - [X] a Feature Request
  - [ ] a Suggestion
  - [ ] a Question
  - [ ] I need help

### LINKED ISSUE #82: Format native blocks - Block Category: Media
Stemming from a block planning conversation between the KE team and the Engineering team, we came up with the following punch list of stuff to do regarding the native blocks that ship with WordPress. Here's what we determined:

### Block Category: Media
Keep: image, audio, video, media & text. All work great out of the box as long as they are wrapped within some kind of `col` class for keeping within Bootstrap.

We decided that we wanted to remove the following blocks:

- **Cover.** We’ll have other blocks that do this in ways that are more fully defined by the UDS system.
- **File:** Remove it in favor of just a native link to the media library. (Consists of the file name + a predefined button.)

We also noted that the default gallery option could use a little additional styling. It currently produces a pretty awful looking grid of square images. (That's been the default gallery styling in WP for a long time.) At the very least, maybe a grid of circle images?

### LINKED ISSUE #83: Format native blocks - Block Category: Design
Stemming from a block planning conversation between the KE team and the Engineering team, we came up with the following punch list of stuff to do regarding the native blocks that ship with WordPress. Here's what we determined:

### Button Block

- Creating an ACF pro block to exactly replicate the Bootstrap 4 button would not be that hard to accomplish.
- An ACF block could also provide support for buttons with alternative markup. (`<button>` v.s. `<a class="btn">`)
- In contrast, manipulating the native block may require a lot of work but would be one less requirement based in ACF.

We decided to attempt the button block manipulation first. Then if that turned out to be problematic, we could turn to ACF. (@wmcconnell has a ACF button block already defined, I think.)

Manipulating the button block would involve:

- Changing the default markup of the button to speak Bootstrap.
- Removing the native outline and fill styles.
- Removing the foreground color choices.
- Removing the option for the border-radius setting.
- Limiting the background colors to black, gold and maroon. (That should also control the corresponding text color.)
- Adding back functionality for the button size. That is probably best done by adding back additional button styles for sm, md and lg.

### Separator
We captured less detail about this. But the default separator block in WP produces an `<hr>` with specific classes for WP. We can possibly manipulate the classes or native styles to produce the UDS defined separators.

### Other Kept items

- **Group**
- **Spacer**
- **Page break**, native functionality for WP pagination
- **More**, native functionality

### Disable list
Disable: Columns ( in favor of bootstrap rows/columns )

### LINKED ISSUE #85: Apply UDS-Bootstrap editor styles & misc theme wide functions
To enable our end users to build blocks with confidence, we should consider how to load the UDS-Boostrap 4 library within the Gutenberg editor. We should also enforce some controls that can be applied globally to all blocks across the theme.

## Editor Styles

The answer here might just be to reference the compiled CSS for the front end of the theme and call it a day. :-)

However, there might be some challenges with the following:

- On a traditional laptop, there may not be enough space in the Gutenberg editing window to see the entire 1200px width of page layout. Using my MBP, I measured the width of `#edit-post-visual-editor` and got a value >1300 as long as the left WordPress menu was closed. Mileage may vary according to the machine, monitor, etc.
- We'll want to further limit the width of the content window for the default page template down to `col-8` size to give a better representation of when text will wrap, etc.

## Other Global Styles

- Remove theme support for align-wide, align-full. Inside of a Boostrap 4 context, these classes will likely just not work.
- Add color palette definitions for UDS approved color choices. 
- Possibly disable custom colors to prevent end-users from going color crazy in places. (Still doable via css and an added class.)
- Disable text size definitions in most places. Those are pretty well defined via code, we could generally not want to allow an end user to increase the font size of something without good reason.

SteveRyan-ASU:
This pull request is intended to close out a few maintenance "chores" within the theme. Here's the list:

- [x] Deregister identified unneeded blocks from the media section of the block inserter. (Closes #82).
- [x] Ditto for blocks from the layout section of the block inserter. (Closes #83)
- [x] Reorganize some of the SASS files, eliminating unneeded stylesheets.
- [x] Cleans up templates other than `page.php` for more consistent markup. (Closes #58)

This pull request also establishes some default block editor options that are important to have to support native blocks in the best way possible. (The following actions will close #85 as well.)

- [x] Disables custom font sizes from all native blocks.
- [x] Disables custom colors from the native color picker.
- [x] Disables custom gradients from the native color picker.
- [x] Adds the ASU native color palette back to the color picker and provides support for the resulting native WP CSS classes.

## Visible results and changes ##
Much of the work here is invisible to the end-user. But, the template file changes should all result in text that is neatly contained within a bootstrap container that's native to the template. (No more extreme-left justified text!)

## To Do ## 
There's still a lot of work to be done in terms of a design for the native `single.php` and `archive.php` templates, both of which are typically used by sites that feature a native blog or native newsfeed. The restructuring on each of those templates was intended to establish a baseline markup for future designs.

Calling attention In particular to `single.php`:

- This template file now features a container that's "baked in" to the layout. 
- The idea here is that eventually, we will probably want to create an editing experience that doesn't require the end-user to need to consider how to use Bootstrap layout blocks in order to create a blog post.
- Eventually, that work will encompass limiting the available blocks in the picker to only the blocks that make sense for this application. (Headings, paragraphs, bullets, images... but no containers/rows or other layout options.)


wmcconnell (approved):
Looks good. Thanks for taking on some of the less glamorous items! Approving and merging into develop.

SteveRyan-ASU:
Addressed merge conflicts, mostly stemming from an add/add situation with `admin.sass`. This PR reorganizes all of the really tiny partials into one well-commented SASS file at the main source. A bit cleaner than our original approach and I don't think we'll need to be adding thousands of lines of CSS to the main admin panel.

wmcconnell (approved):
Okay, second time around: looks good, and the conflicts are resolved. I think you're right about merging lots of little admin styles into one file. Approving, as we want to get the latest into dev and bump the version for our work on the RTO site.

PR #131, opened 2021-02-24
Replace Dummy Links in Header

wmcconnell:
Replacing two placeholder links in `header.php` that were just links to the top of the page. Suggested changes are:

* Update the main ASU logo in the header to link to 'https://www.asu.edu'
* Update the 'Sign In' link in the global menu bar to link to 'https://my.asu.edu'

We have been asked if the 'Sign In' link could also pass users through to the WordPress site, but that will require some investigation. This fix at least means that an unauthorized user who clicks on 'Sign In' will be logged in to their ASU account via CAS - which is the real purpose of that link.

### DIFF
--- a/header.php
+++ b/header.php
@@ -113,7 +113,7 @@ if ( ! empty( $c_options['hotjar_site_id'] ) ) {
 							<a class="nav-link" href="https://my.asu.edu">My ASU</a>
 							<a class="nav-link" href="https://asu.edu/about/colleges-and-schools">Colleges and Schools</a>
 							<div class="nav-link login-status">
-								<a class="signin" href="#">Sign In</a>
+								<a class="signin" href="https://my.asu.edu">Sign In</a>
 							</div>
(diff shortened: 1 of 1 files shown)


SteveRyan-ASU (approved):
Both of these links look like good choices in terms of what is supposed to be there according to the brand standard. 

- The Preact version of the header has a hard coded link to asu.edu from the vertical logo in the desktop presentation. So that aligns with what we have here.
- The Bootstrap version has no link assigned, just `href="#"` and that's it. Totally fine to encode that as a link back to the mothership, I think.

An alternate destination for the sign-in link would be to head to {url}/wp-admin. That would then redirect to the login page if the user wasn't logged in. But, I think the intent of those links is to determine if you are signed in to the ASU ecosystem as a whole. So pointing back to my.asu.edu makes sense as well.

nathan-rollins (approved):

Issue #86, opened 2021-01-20
Create block: UDS-Bootstrap "Container"


SteveRyan-ASU:
In order for us to use the set of native blocks that ship with WordPress within the context of a Bootstrap enabled page, we will need to create a block that can function as a combination of a Boostrap container, BS row and BS column-set. For planning purposes, we should be thinking about a block that produces the following structure:
```
<section id="unique" class="bg network-white">
   <div class="container">
      <div class="row">
         <div class="col-md-6">
...
```
It's looking like we may want to have the following options available within this block:

**Section**
- Dropdown menu for all available UDS background textures.
- Media file control to upload/select a background texture from the Library. (Would produce a `style="background-image:url"` style code when enabled.

**Container**
- Controls for what kind of container you want to use. (Where is the breakpoint? XL, LG, MD, etc.)

**Column**
- Some kind of control for common layout patterns. Would be great if we could mimic what @zainabD origionally did with the sidebar panel. Radio buttons under icons kinda thing. But, simpler is sometimes better and easier to execute, so even just a dropdown with the descriptions is a good start.
- Offhand, I'd say that those patterns should be:
  - col-4, col-8 (sidebar left)
  - col-8, col-4 (sidebar right)
  - col-3 x 4 (four columns)
  - col-4 x 3 (three columns)
  - col-6 x 2 (two columns)

SteveRyan-ASU:
In thinking through this story, we may also want to do some fancy work with ACF to make the following happen:
- Make the control for the column pattern an optional control.
- When it's enabled, choose a pattern and drop in as many `<InnerBlocks />` tags as are needed.
- When it's disabled, don't produce the `<div class="col-6">` markup at all. Instead produce one `<InnerBlocks />` tag and limit the contents of that tag to just the custom HTML element. That would be the "roll your own columns" kind of approach... but all interior content would also have to be hand coded as well.

PR #132, opened 2021-02-26
UDS Background Section

SteveRyan-ASU:
This block establishes an optional `<section>` element exclusively to wrap a standard Bootstrap 4 container. The purpose of this wrapper is to:

- Provide a consistent semantic HTML5 markup for the page.
- Provide a consistent class to use to reinforce the max-width:1920px rule.
- Provide a block interface to select approved ASU background textures or approved colors from a list.
- Provide a block interface to allow users to upload their own background texture.
- Automatically set the margin and padding rules for this element in the context of surrounding elements.
- Automatically set the background color and text color on the interior container based on the supplied background texture.

The output of the block is:
```
<section class="uds-section ..." >
...
</section>
```
...where `uds-section` is followed by a series of optional utility classes. Other notes:

- It allows only two blocks as inner content: the Bootstrap container block and the custom HTML block.
- It also comes with sample content that pre-populates the `<section>` with an interior `<container>` when you first add the block to the editor.


SteveRyan-ASU:
Controls for the block via ACF. (Radio buttons all the way down.)
image

SteveRyan-ASU:
A note on the usability of this block within the editor:

- The interface for using this block as the starting point for a new section still feels very clunky.
- It's relatively easy to set up the remaining BS row and columns incorrectly. The background colors and textures make it difficult to see where to click next for the row and columns.

It works well enough out of the box, but a better use case for this would be to establish a block pattern that added this block + the container/row/columns for common layouts.

SteveRyan-ASU:
If this block is merged it would close #86. 

While not exactly what was prescribed in that issue, this block + the other blocks added by the container block plugin get us to the intended end result.

SteveRyan-ASU:
One additional note: I was hoping that we would be able to drag a block (or a series of blocks) into this block like you can with the native media/text block (as an example.) But, I was unable to figure out how to do that within the ACF context.

zainabD (changes requested):
Good work Steve!, I noticed some issues when using the block with different combinations, also added some suggestions below:

- [ ] [Issue] The Additional CSS class(es) field is not connected.
- [Suggestion] We need to add the option to change the container's background to transparent, as in some cases we need to add text, header or button over the background section without having to have another colored background. 
image

- [Question] Why can't we get the opacity options when selecting the UDS background texture, while these 3 options become available when uploading an image?
image


- [ ] [Issue] When I check the **Fluid** option on the container block, the container background becomes transparent, since you are targeting the **.uds-section.bg.topo-black .container**, not .**container-fluid**. 
- [ ] [Issue] In dashboard view, when checking the **Fluid** option, the container block remains the same with background color as you're targeting .**editor-styles-wrapper .uds-section.bg.topo-black .wp-block-wp-bootstrap-blocks-container** .
On page view: 
image

On admin view: 
image



- [ ] [Issue] if I selected a dark bg, all the text inside will be white by default per this css: 

```
editor-styles-wrapper .uds-section.bg.morse-code-black .wp-block-wp-bootstrap-blocks-container, .editor-styles-wrapper .uds-section.bg.network-black .wp-block-wp-bootstrap-blocks-container, .editor-styles-wrapper .uds-section.bg.topo-white .wp-block-wp-bootstrap-blocks-container {
background-color: #191919;
    color: #fff;
}
```

In this case I created a **background section** with **dark** option, then inside that section, created another **background section** with **uploaded image** and **UDS White** option for the container background, but the text was still showing in white as it's a child of the first dark background section. 
I think this should only target the direct child. 
image

- [ ] [Issue] The topographic map - white and dark options need to be switched. 
Screen Shot 2021-03-01 at 2 43 11 PM

SteveRyan-ASU:
Picking this back up and hopefully pushing it across the line. 

It's been a while since I last looked at this, but to get it back up to speed, I started with the reorganization of some of the ACF fields and simplified the controls a little bit.

The block still has three main "modes" which are triggered by the opening radio button for "pattern or color"
### UDS Background Texture ###
- Produces a background section with an approved pattern.
- Styles the inner container minimally and appropriately with either a white or UDS-Gray-7 color.
- Adds `text-white` class to the `<section>` element to assist with dark on dark text.

### ASU Background Color ###
- Presents a choice for a background color as of of the "official, supported" options: Gray 1, Gray 2 or Gray 7.
- Allows for a fourth choice to "choose my own color" with a sufficient warning about allowable colors from the palette.
- A color picker is presented to the user conditionally to allow for a custom color.

### Upload a file ###
- Presents a dialog for uploading a file into the media library.
- Conditionally presents an additional accordion of options pertaining to the CSS background style.
- Allows for a background-position and background-repeat option. Those two options handle the bulk of the difference between a photo background (no-repeat, cover) and a background pattern (repeat).
- Uploaded files also allow for the user to manually select an inner background color from the approved set of colors. (White, Gray 1, 2 and 7. Also allows for a no inner background color option.)

Addressing some of the concerns from @zainabD's last review:

- All styles applied to the inner "container" for this block are now targeting a class of [class*="container"] as opposed to just `.container`. The wildcard selector should cover all container types including fluid and responsive-fluid.
- Any style applied to the inner container from a class within the `section` can immediately be overridden by a custom CSS field applied to the container directly. 
- I added the utility code to allow for Additional CSS rules to be applied to this new block.
- The inner blocks feature of this block does not allow for a section to be nested within another section. So, via ordinary means, we shouldn't run into problems where the inner container rules need to be applied to only a direct child of the section block.

The opacity options that were initially within this block were dropped for simplicity. The best way to implement these back would be to allow the user to pick a custom inner container color via the color picker. (The output of the color picker can be presented as an RGBA value, so you get a color and an opacity in one shot.)

css/admin.css
```diff
@@ -91,10 +94,14 @@ Enqueued only within the admin screen. (Via enqueue.php)
...
...pdGVtLWhhbmRsZSAuaXRlbS10aXRsZSB7XG4gICAgbWFyZ2luLXJpZ2h0OiA1ZW07IH1cblxuLm1lbnUtaXRlbS1kZXB0aC0xMiAubWVudS1pdGVtLWJhciAubWVudS1pdGVtLWhhbmRsZSB7XG4gIGJhY2tncm91bmQtY29sb3I6ICNGOTk7IH1cbiAgLm1lbnUtaXRlbS1kZXB0aC0xMiAubWVudS1pdGVtLWJhciAubWVudS1pdGVtLWhhbmRsZSAuaXMtc3VibWVudTo6YWZ0ZXIge1xuICAgIGNvbnRlbnQ6ICcgLSB0b28gZGVlcCEnOyB9XG4gIC5tZW51LWl0ZW0tZGVwdGgtMTIgLm1lbnUtaXRlbS1iYXIgLm1lbnUtaXRlbS1oYW5kbGUgLml0ZW0tdGl0bGUge1xuICAgIG1hcmdpbi1yaWdodDogNWVtOyB9XG5cbi5tZW51LWl0ZW0tZGVwdGgtMTMgLm1lbnUtaXRlbS1iYXIgLm1lbnUtaXRlbS1oYW5kbGUge1xuICBiYWNrZ3JvdW5kLWNvbG9yOiAjRjk5OyB9XG4gIC5tZW51LWl0ZW0tZGVwdGgtMTMgLm1lbnUtaXRlbS1iYXIgLm1lbnUtaXRlbS1oYW5kbGUgLmlzLXN1Ym1lbnU6OmFmdGVyIHtcbiAgICBjb250ZW50OiAnIC0gdG9vIGRlZXAhJzsgfVxuICAubWVudS1pdGVtLWRlcHRoLTEzIC5tZW51LWl0ZW0tYmFyIC5tZW51LWl0ZW0taGFuZGxlIC5pdGVtLXRpdGxlIHtcbiAgICBtYXJnaW4tcmlnaHQ6IDVlbTsgfVxuXG4vKi0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS1cbjIuIEZvbnQgQXdlc29tZSBmaXhlcyBmb3IgR3V0ZW5iZXJnXG4tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLS0tLSovXG4uZWRpdG9yLXN0eWxlcy13cmFwcGVyIHN2Zy5zdmctaW5saW5lLS1mYSB7XG4gIHZlcnRpY2FsLWFsaWduOiAtLjEyNXJlbTsgfVxuIl19 */
```
zainabD, line 98: It looks like this wasn't merged correctly, why do we still have **<<<<<<< HEAD** and **>>>>>>> develop**

css/admin.css
```diff
@@ -91,10 +94,14 @@ Enqueued only within the admin screen. (Via enqueue.php)
...
+>>>>>>> develop
```
zainabD, line 107: Same here ... 
I noticed this because I'm getting the following error on console: 
`Error while traversing the CSS: Error: undefined:7:1: missing '{'`
and the editor styling are broken

css/admin.min.css
```diff
@@ -1,2 +1,6 @@
...
+>>>>>>> develop
 /*# sourceMappingURL=admin.min.css.map */
```
zainabD, line 6: same here, conflicts were not resolved when merged develop into this ..

css/admin.min.css.map
```diff
@@ -1 +1,5 @@
...
+>>>>>>> develop
```
zainabD, line 5: Same issue

css/theme.min.css
```diff
@@ -1,2 +1,6 @@
...
...4{margin-top:-1.5rem!important}.mr-lg-n4,.mx-lg-n4{margin-right:-1.5rem!important}.mb-lg-n4,.my-lg-n4{margin-bottom:-1.5rem!important}.ml-lg-n4,.mx-lg-n4{margin-left:-1.5rem!important}.m-lg-n5{margin:-3rem!important}.mt-lg-n5,.my-lg-n5{margin-top:-3rem!important}.mr-lg-n5,.mx-lg-n5{margin-right:-3rem!important}.mb-lg-n5,.my-lg-n5{margin-bottom:-3rem!important}.ml-lg-n5,.mx-lg-n5{margin-left:-3rem!important}.m-lg-n0{margin:0!important}.mt-lg-n0,.my-lg-n0{margin-top:0!important}.mr-lg-n0,.mx-lg-n0{margin-right:0!important}.mb-lg-n0,.my-lg-n0{margin-bottom:0!important}.ml-lg-n0,.mx-lg-n0{margin-left:0!important}.m-lg-n1{margin:-.5rem!important}.mt-lg-n1,.my-lg-n1{margin-top:-.5rem!important}.mr-lg-n1,.mx-lg-n1{margin-right:-.5rem!important}.mb-lg-n1,.my-lg-n1{margin-bottom:-.5rem!important}.ml-lg-n1,.mx-lg-n1{margin-left:-.5rem!important}.m-lg-n2{margin:-1rem!important}.mt-lg-n2,.my-lg-n2{margin-top:-1rem!important}.mr-lg-n2,.mx-lg-n2{margin-right:-1rem!important}.mb-lg-n2,.my-lg-n2{margin-bottom:-1rem!important}.ml-lg-n2,.mx-lg-n2{margin-left:-1rem!important}.m-lg-n3{margin:-1.5rem!important}.mt-lg-n3,.my-lg-n3{margin-top:-1.5rem!important}.mr-lg-n3,.mx-lg-n3{margin-right:-1.5rem!imp
```
zainabD, line 6: <<<<<<< HEAD and >>>>>>> develop

inc/uds-blocks.php
```diff
@@ -53,6 +53,7 @@ function my_acf_blocks_init() {
 			'/content-sections', // Miscellaneous content sections.
 			'/headings', // A UDS Headings block.
 			'/program-card', // UDS Program Cards.
+			'/background-section', // UDS Background section.
```
zainabD, line 56: I added this line since the block was not registered anywhere.

zainabD (changes requested):

templates-blocks/background-section/background-section.php
```diff
@@ -0,0 +1,74 @@
...
+$upload = get_field( 'uds_background_section_upload_url' );
+$innercolor = get_field( 'uds_background_inner_color' );
+$repeat = get_field( 'uds_background_image_repeat' );
+$position = get_field( 'uds_background_image_position' );
+
+// Retrieve additional classes from the 'advanced' field in the editor.
+if ( ! empty( $block['className'] ) ) {
+	$additional_classes = $block['className'];
+}
+
```
zainabD, line 21: Can you give the $additional_classes an empty value if it's empty to avoid the following error: 
`Notice: Undefined variable: additional_classes`

zainabD (changes requested):
Over all it looks good, I just added a note to fix the error "`Notice: Undefined variable: additional_classes`"
- Also, I have a couple of suggestions: 

- [ ] In image background case, can we add an additional option beside the background-position and repeat, a new option to control the **background size** whether it's contain or cover? 

- [ ] Is it possible to add the option of curve backgrounds, like the ones show on SFIS and SOS xd mockup: 

image
 
We added these to SFIS and SOS home page an additional css as shown below (https://sfis.qa.rtd.asu.edu/): 

```
    border-bottom-left-radius: 55% 10%;
    border-bottom-right-radius: 55% 10%;
    border-top-left-radius: 55% 10%;
    border-top-right-radius: 55% 10%;
    width: 1900px;
    background-position: center;
    position: relative;
    left: 50%;
    transform: translateX(-50%);
```

zainabD:
One more thing I forgot to mention: 
In SFIS and SOS, we have use a background image that is only repeats on x, like this one: 
image

Can we add this option under background repeat, to choose whether to repeat on x or on y?

templates-blocks/background-section/background-section.php
SteveRyan-ASU: Resolved. Good catch. :-)

SteveRyan-ASU:
Requested changes in the latest push to this branch:

- The choice of the inner text color is now a user-driven choice rather than something that was applied at the CSS level with the choice of "approved" background textures. The control defaults to normal text until you explicitly set it as white. 
- Added support for both repeat-x and repeat-y as `background-repeat` options. The options of `space` and `round` were omitted to create a better user experience, although adding them back would be trivial if desired.
- Added explicit support for the `background-size` options of cover and contain. I think it would be possible to add a conditional control within the block that would allow the user to define a string for an explicit numeric value (`30% auto`, for example). But, that is also possible to do by adding a custom CSS class and rolling the rule out that way.

Adding these choices as completely user-configurable options DOES allow the block to produce some code that wouldn't normally make sense. (`background-size:cover` does not require an explicit `no-repeat` value, for example.) But, the produced code is still valid markup

I also added the oval background shape as a utility class to this element, but I left the feature without an explicit control within the UI to turn it on or off. 
- Add the class of `.shape-oval` to the advanced options area to trigger the result.
- Unfortunately, with this method of delivering this feature, the result is not visible within Gutenberg.
- Plays well with all of the other options within the block including the "roll your own background image" and background colors. Also respects the `top/middle/bottom` rules of the background-position controls within the block.

zainabD (approved):
Nice and clean work! good job Steve.
I tested this locally and everything seems to be working with no issues.

wmcconnell (commented):
I don't know if it was just me, but the custom background image was not working as I expected. I chose an image (a repeatable pattern, 240px wide) as my background and got this:

Screen Shot 2021-05-06 at 3 28 37 PM

The small image is stretched, and not repeating. The final CSS seems to have an unwanted character in there, a `\` between `center` and `cover`:

```<section class="uds-section media-file inner-bg-white " style="background: url(http://mythemedev.local/wp-content/uploads/2021/05/wildtextures_brick-elevation-tiles-seamless-texture-e1498137214248-480x461-1.jpg) center / cover repeat;">```

I'm not even sure if you want both `center` and `cover`. Removing `/ cover` fixes the issue, with the image actually repeating and looking like this:

Screen Shot 2021-05-06 at 3 32 06 PM

Haven't figured out where the `/` comes from, but could it be taking the place of an actual parameter and causing some issues with the repeat setting?

SteveRyan-ASU:
Good catch @wmcconnell. Lucky for me the fix was small. 

- Added "auto" as a background-size option within ACF and set it as the default.

**Details**

The `/` within the `background:` inline CSS rule is part of the spec. It's a required value when setting both a background position and size when using the shorthand rule set. (https://developer.mozilla.org/en-US/docs/Web/CSS/background)

Removing the `/ cover` or `/ contain` declarations from the rule still produces valid markup that is the equivalent of `/ auto` for the background-size. 

I didn't catch the difference visually between "contain" and "auto" because the tile I was using to test was too small to notice the difference. Here's the difference:

- Both "contain" and "auto" when used with either of the repeat settings will produce a tiled background image.
- Contain will have the effect of scaling down the tiles so that none of the tiles are cut off from the size of the background area. (It will repeat an exact whole number of times within the given area.)
- The "auto" value does what you would expect - repeats and cuts off the excess of the image based on the background-position.

wmcconnell (approved):
Nice work, Steve! Easy to create a series of page sections with various background styles and with the proper UDS spacing built in. I was able to quickly fill a page with multiple sections containing images, text, and cards and have it look great with minimal effort.

PR #136, opened 2021-03-01
Card Block

### LINKED ISSUE #95: ACF Block: Card(s)
We need to implement a UDS card as a block. Rather than put all the requirements here, see [the XD file entry for cards](https://xd.adobe.com/view/56f6cb78-9af5-4b12-b4ce-ef319f71113f-03a5/screen/10657514-1129-405c-9726-23ac122ae4bf/), [this CodePen](https://codepen.io/asudigital/pen/jOWPNWK), or [the Storybook](https://unity.web.asu.edu/@asu-design-system/bootstrap4-theme/?path=/story/components-cards--default-card).

If creating a single, configurable block that can handle all card types (basic, news, events, story) ends up being too complicated, or creates a bad user experience, we should consider breaking this down into multiple blocks - one per card type/style.

wmcconnell:
Resolves #95

After many updates and changes, some last-minute feature additions, and dozens of cards made during the process, it's finally time to add this card block to the theme.

- A new ACF Pro block for creating UDS cards. Features include:
    - Choose from multiple UDS card styles: basic, event, and story
        - Conditionally hide/show event input fields based on type 
    - Choose from multiple header styles: none, icon, or image
        - Conditionally show image and icon inputs based on type 
    - Standard vertical, or wider horizontal cards (and switch between on the fly)
    - Add multiple buttons, links, or tags with ACF Pro repeater fields
    - Apply additional classes, when needed, with the 'Advanced' input in Gutenberg
    - Create 'hover' cards
        - Conditionally show/hide button, link, and tag inputs based on choice
        - Conditionally show/hide actual buttons, links, and tags on the display end based on choice

This should cover nearly all of our _static_ card needs, as far as I can see, and I've had few issues with them (the card styles were put together much better than the block quotes).

There are a few things that **are not** included here, both recent additions to the XD file:

- Media type overlays: These icons indicate that a card links out to a specific type of media. I need some help figuring out the best way to implement these without changing the existing markup, if that's even possible.
- Checkbox overlays: checkboxes for selecting multiple cards in order to drive some kind of interaction. This seems to go beyond a simple static card, but we could potentially add this with the ACF JavaScript API.

Here are some (small) screen shots:

1) Eight different card types/styles made with the block:

Screen Shot 2021-03-01 at 2 20 18 PM

2) A basic UDS card inside Steve's content/image overlay block

Screen Shot 2021-03-01 at 2 38 51 PM

3) A simple, stylized grid, with background classes applied to the cards

Screen Shot 2021-03-01 at 3 11 24 PM


wmcconnell:
After submitting this, I noticed I hadn't pushed my latest commits! They've been added.

PR #138, opened 2021-03-04
added theme.min.css as an editor style to fix the style loading issue…

zainabD:
I added the line :
**add_editor_style('css/theme.min.css');** 
after 
**add_action( 'after_setup_theme', 'uds_wp_gutenberg_css' );**
to fix the style loading issue on admin dashboard.

### DIFF
--- a/inc/enqueue.php
+++ b/inc/enqueue.php
@@ -65,7 +65,7 @@ if ( ! function_exists( 'uds_wp_gutenberg_css' ) ) {
 	 */
 	function uds_wp_gutenberg_css() {
 		add_theme_support( 'editor-styles' );
-		add_editor_style( get_template_directory_uri() . '/css/theme.min.css' );
+		add_editor_style( 'css/theme.min.css' );
 	}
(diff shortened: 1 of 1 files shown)


wmcconnell (approved):
Thanks, Zainab! We have tested this on both a broken local site (my copy of the QuickStart) and a live sandbox dev site on WP Engine - problem solved on both. Also applied the change to my working local site with no obvious side effects.

Approved.

nathan-rollins (commented):
There is a newer function (introduced in WP 4.7.0) for loading theme files: https://developer.wordpress.org/reference/functions/get_theme_file_uri/

Maybe we should be using it instead? It is designed to allow child themes to override, if needed.

zainabD:
> There is a newer function (introduced in WP 4.7.0) for loading theme files: https://developer.wordpress.org/reference/functions/get_theme_file_uri/
> 
> Maybe we should be using it instead? It is designed to allow child themes to override, if needed.

I just tried the new function get_theme_file_uri (), and it didn't work, we're getting the same errors in the inspect.

PR #139, opened 2021-03-05
Gulp browserSync function

zainabD:
I updated the gulpfile.js file, added .**pipe(browserSync.reload({ stream: true }))** to end of each: 
**minifycss** and **scripts** tasks to make the live stream work when changes are made.

- This will allow us to view changes right after we save the sass file as shown in the video bellow:

https://user-images.githubusercontent.com/15880606/110175609-4da7cc00-7dbf-11eb-9638-bfc8c1e3e249.mov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ Thumbs.db
 phpcs.xml.dist
 .phpcs.xml
 .phpcs-cache
+
+browserSyncOptions.json
--- a/browserSyncOptions.json
+++ b/browserSyncOptions.json
@@ -1,0 +1,7 @@
+{
+	"browserSyncOptions": {
+		"proxy": "localhost:8080/",
+		"notify": false,
+		"files": ["./css/*.min.css", "./js/*.min.js", "./**/*.php"]
+	}
+}
--- a/gulpconfig.json
+++ b/gulpconfig.json
@@ -1,9 +1,4 @@
 {
-	"browserSyncOptions": {
-		"proxy": "localhost:8080/",
-		"notify": false,
-		"files": ["./css/*.min.css", "./js/*.min.js", "./**/*.php"]
-	},
 	"paths": {
 		"js": "./js",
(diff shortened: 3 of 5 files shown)


wmcconnell (changes requested):
I don't know what I did trying to test this out, but it broke my local setup. I cannot get this to actually work. I think it has to do with my particular configuration, and maybe we can get together tomorrow and go over your settings.

Regardless of that, I think we might want to add this to the `package.json` file along with the other NPM scripts. See line 8 of `package.json` for the current scripts. You should be able to add something like:

`"sync": "gulp watch-bs"` (where 'sync' is whatever you want to call the command) so that someone without Gulp installed globally could run `npm run sync` and use the Gulp that's built in to the project.

zainabD:
As mentioned in README file, in order to run the browser sync you need to change the proxy option in json file: 

> BroswerSync
> Or, to run with Browser-Sync:
> 
> First change the browser-sync options to reflect your environment in the file /gulpconfig.json in the beginning of the file:
> {
>     "browserSyncOptions" : {
>         "proxy": "localhost/theme_test/", // <----- CHANGE HERE
>         "notify": false
>     },
>     ...
> };
> then run: $ gulp watch-bs

- Per Walt suggestion, I created a separate json file (**browserSyncOptions.json**) for browser sync options only, and this file could be added to gitignore file since it has to be updated per each individual's local settings.

- I also added the line **"sync": "gulp watch-bs"** in package.js file to make the browser sync function works in npm by running **npm run sync** instead of **gulp watch-bs**

I tested it locally and it works with no issues.

wmcconnell (approved):
Looks good! I think it was a good idea to move the browser sync options to their own file, and then not track that file. I was able to set up my local system with no problem this time, and everything worked - and quickly! I even had the site on my desktop and phone, and watched them both update nearly instantly after saving. Even the synced actions, like scrolling, worked.

PR #145, opened 2021-03-11
Remove Core Block Patterns

wmcconnell:
If we're going to build a library of UDS block patterns, we should probably remove the core patterns - which may use core blocks that we have removed (or are planning to remove).

- Update our `uds-blocks.php` file to remove support for core block patterns. This file is where we also remove core blocks, so it made sense putting this here rather than in a new file.


inc/uds-blocks.php
```diff
@@ -80,3 +80,11 @@ function uds_wordpress_unregister_native_blocks() {
 
 	add_filter( 'allowed_block_types', 'uds_wordpress_unregister_native_blocks' );
 }
+
+/**
+ * Deregister core block patterns
+ */
+function remove_core_patterns() {
```
zainabD, line 87: I would add the line before starting the function: 
if ( ! function_exists( 'remove_core_patterns' ) )  
To avoid getting errors in case the function name was used in a child theme.

zainabD (changes requested):

zainabD (approved):

PR #148, opened 2021-03-12
border-[breakpoint]-side classes

zainabD:
I added new border class to make it possible to use bootstrap breakpoints with border-[side] classes:
Using the format: .border-{breakpoint}-{side}-{size} [size works on 0 only)]
for example
.border-lg-top || .border-lg-top-0
.border-lg-right || .border-lg-right-0
.border-lg-bottom || .border-lg-bottom-0
.border-lg-left || .border-lg-left-0
.border-lg-x **for both left and right**
.border-lg-Y **for both top and bottom**
And so on ...

Files I updated: 
css/theme.css
css/theme.min.css
css/theme.min.css.map
sass/theme/_theme.scss

A good use of these classes: 
we only needed these border -right to show on lg and md, but not on sm and xs, as well as the border-bottom: 
image

image


SteveRyan-ASU (dismissed):
Good work Zaniab. I'm OK with including these changes as they are, but I had a possible suggestion.

Given this solution is something that we will consider keeping around for the longer term, I'm wondering if you would consider keeping the SASS code in its own partial (`theme/_responsive-borders.scss` maybe?) rather than in the more generic "catch-all" file at `_theme.scss`?

zainabD:
Per Steve's suggestion, I added a new sass file called responsive-borders and imported it in the mail theme.scss

wmcconnell (approved):
Looks good, and works as advertised in my testing. Zainab added the ability to use just 'border' (along with responsive variants), which only makes it better. Approving.

PR #150, opened 2021-03-16
Alert Block and Bootstrap Updates

### LINKED ISSUE #98: ACF Blocks: Alerts
We need to implement a block for static alerts [as shown here in Storybook](https://unity.web.asu.edu/@asu-design-system/bootstrap4-theme/?path=/story/components-alerts--alerts). The user should be able to:

- choose the style (note that each style uses a particular icon)
- enter the text
- turn the 'x' button on or off

While alerts would usually be created by the system, after some operation has succeeded/failed, this block would be for creating specific, static messages to place at the top of a page.

earlalex:
Resolves #98

Provides a UDS Alert block, with updated styling from the latest upstream Bootstrap theme. As a result of pulling the latest Bootstrap in order to get alert updates, this branch also includes some unrelated theme updates. Rather than tease them out into separate PR's, we are including everything here.

Changes requested here, then, include:

- A new alert block, with updated 'simple' toolbar for the WYSIWYG editor, ability to hide the icons, and new styles from upstream that fix the close button alignment and opacity issues.
- Removal of our specific SASS files for the Image-Overlap block. Those styles are now part of the UDS Bootstrap theme. Deleted our local file and removed the include from `theme.scss`
- Any and all updates to the UDS Bootstrap theme since the last time we pulled the latest. Most of these are smaller styling adjustments. Notable changes include:
    - The inclusion of the image-overlap styles (as noted above)
    - Updated styling for tabs and tab content areas, most of which are related to the mobile sizes.
    - Buttons no longer expand on hover at mobile screen sizes
    - The renaming of the `quick-links` class to `uds-quick-links`, fixing a conflict with the native WordPress admin `quick-links` class


wmcconnell:
Updated branch now uses the built-in Alert functionality of Bootstrap - including fading out when closed.

zainabD:
Good job Earl and Walt, I just noticed few things here: 
- [ ] The alert icon show on the top of the box, while in the xd it seems to be vertically at the middle: 

image

image

- [ ] Also the space to the left of the "x" button on mobile doesn't look right: 
image

- [ ] The additional CSS class field is not connected to the markup: 
image

Can we make sure that none of these updates will break our current websites like **devils drop off** and **RTO** ..

zainabD (changes requested):
I tested this on RTO locally, it looks good, didn't break anything. 

- [ ] The **Additional css classes** field still needs to be connected. 
- [ ] and I think that the "x" button need to stay on the corner. 

Good job on moving the icon to always be at the middle.

zainabD (approved):

Issue #152, opened 2021-03-18
Code Editor Display Width


jehollow:
When using the Code editor in the options section it maxes out the edit post text editor body to the point you can no longer exit the code editor and go back to the visual editor. Below is the class that is effecting the width.

.edit-post-text-editor__body {
    max-width: none !important;
    margin-left: 5%;
    margin-right: 5%;
}

nathan-rollins:
That is from an editor tweak I added way back in the beginning to make the editor full-width. The function that applies that CSS is located in 'inc/editor.php', function gutenberg_editor_full_width():

https://github.com/asu-ke-web-services/UDS-WordPress-Theme/blob/8c1fdf809c2c4a4e379faa0e3a8fa79ed22e0e70/inc/editor.php#L90

If you disable that function, then the editor will return to its original size. Depending on how this function is causing your issues, maybe you just need to tweak the CSS properties? Otherwise, if the team agrees that you prefer the original editor size, feel free to remove the function.

nathan-rollins:
And now I found a note in the WP Theme Developers Guide about changing the the default width of the Gutenberg Editor. This looks like it might be a better solution than what I used above.

https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#changing-the-width-of-the-editor

zainabD:
Removing this part should fix the issue: 
.edit-post-text-editor__body { 
 		max-width: none !important; 
 		margin-left: 5%; 
 		margin-right: 5%; 
 	} 

- To apply the wordpress solution that Nathan mentioned, we need to add a new attribute [data-align="full"]

zainabD:
This was fixed in this PR: [Fix the gutenberg_editor_full_width issue #153]

PR #153, opened 2021-03-19
Fix the gutenberg_editor_full_width issue

zainabD:
This is a fix for issue #152 
I updated the styles under function **gutenberg_editor_full_width** on inc/editor.php page, removed the styles added to .edit-post-text-editor__body: 

```
.edit-post-text-editor__body {
max-width: none !important;
margin-left: 5%;
margin-right: 5%;
}
```


wmcconnell (approved):
This looks good, and solves the ever-expanding code editor problem in my testing. Approved.

PR #154, opened 2021-03-19
Hero <img>

zainabD:
In this branch I updated the hero section, removed the **background-image** style and added an html `` tag. 

- Updated **hero.php** file, removed the background-image style from the uds-hero div, add an  tag with srcset inside the div.
- Updated **theme.sass**, added the gradient background to the pseudo selector of the uds-hero (after).
- Added object- property to the hero img.

Screen Shot 2021-03-19 at 4 10 47 PM


wmcconnell (approved):
This looks good to me, and is working properly on my local. Using an `` tag just makes more sense, and it gets us the Web-P and source set features we're looking for. Approving this so I can use it as a basis for the multiple button version.

PR #155, opened 2021-03-20
Block Enhancement and Cleanup

### LINKED ISSUE #142: Add Support for 'Advanced' Field to Headings and Buttons
Add support for 'advanced' class in UDS-Heading and UDS-Button

wmcconnell:
This branch adds some needed features to some blocks, attempts to create a more consistent experience when using the blocks, and includes some general upkeep in an attempt to make the code and comments similar across the blocks.

### Larger Features
- Add support for additional CSS classes to blocks that didn't have it (especially buttons and headings, as we need that right now). Users can use the 'Additional CSS Classes' input in Gutenberg to add arbitrary CSS classes to our blocks as required. Resolves #142.
- Update the button block to use the ACF 'Link' field instead of a URL field. Previously, you could only use a full URL here, and not a relative path. Users can now select a post or page from the site with full search capability, provide a full URL, or use a relative path.
- Update the Card block to use the ACF 'clone' field type for buttons. Rather than repeat the same input fields for the button section of Cards, it's now linked to the button field group via the clone field and uses the exact same fields. Added support for all those fields in the Card code as well.
- Update the Page Hero field group to allow users to choose a hero image file from the media library

### Minor Features
- Add the ability to add a pre-defined spacing on the top of header tags with a simple yes/no control
- Defaulted our blocks to preview mode (not edit mode) whenever possible, figuring that it was a better user experience
- Moved the 'additional classes' off of the button markup itself, and on to the surrounding `<div>` tag

### ACF
- Renamed our block groups with a `UDS Block: {name}` pattern to keep them together in the field group list, and make them easy to distinguish from other field groups
- Made use of ACF's tabs in some places (block quote, page hero) to organize fields for a better experience
- Updated the preview data on some blocks for better (but not quite what I want) previews when inserting blocks via the block selector

### Misc Cleanup
- Tried to make the comment style similar across our block template files; replace some placeholder comments from the sample blocks
- Removed the sample block field groups, their code, and their associated JSON files. We have enough real examples that the samples are not really needed
- General code cleanup


acf-json/group_5ef954da477f5.json
```diff
@@ -28,30 +44,100 @@
...
             "wrapper": {
-                "width": "60",
+                "width": "",
                 "class": "",
                 "id": ""
             },
             "default_value": "",
             "placeholder": ""
         },
+        {
```
nathan-rollins, line 93: I don't think this Media File should be added in this PR. If we are going to switch, that would be in Zainab's PR that implements everything needed.

nathan-rollins (changes requested):
I don't think the Media File field should be added in this PR. That needs to be a separate PR that also updates the page templates, etc. Zainab is already working on that.

SteveRyan-ASU (approved):
This looks good. I appreciate the attention to detail from an end-user perspective.

- Opening blocks in "preview" mode is the clear winning choice. You get the sense of what the block is supposed to look like + the ability to edit settings in the sidebar all at the same time. 
- ... so much so, that it makes me wonder if our documentation should encourage people to edit all ACF blocks in this way instead of bothering to toggle the preview/edit mode in the toolbar. 
- The use of the ACF Link field for pretty much anything that could possibly be a link to a page within the site also makes a lot of sense. That interface is really familiar to WordPress users and provides lots of flexibility without us having to reinvent the wheel.
- Organizing controls within tab groups is another good improvement. Saves screen real-estate and provides the easiest way to "label" a group of controls that concentrate on one set of user input.

SteveRyan-ASU:
I'll also add that I am behind the times in understanding the coordination of work taking place across multiple PR's. So, I'll let you and @zainabD talk about the potential for a merge conflict with work in a different PR.

acf-json/group_5ef954da477f5.json
wmcconnell: Well, I already messed this one up. I didn't understand what you were saying, as the screen shot I got in the email notification of your comment looked like this, which was somewhat confusing:

Screen Shot 2021-03-22 at 11 55 44 AM

And now I've already merged her branch, intending to go figure this one out later. Looking at the actual comment in context on Github makes more sense and now I see what you're saying.

Possibly ineffective counter-argument: my branch was about cleaning up field groups and associated blocks, where Zainab's was related primarily to swapping out a CSS background image for an actual `` tag, so it's okay that my branch takes care of adding this field to the group and dealing with any conflicts the new field may cause.

zainabD (approved):
Good work Walt! I checked the following updates and they all worked with no issues:
- Additional CSS field in buttons and headings.
- The new link field in button block.
- The updates on Card block.  

> - [I'm just wondering why are the buttons, tags and link doesn't work when the hover option is checked?]
> - [also, why don't the buttons on card have the additional css field?]
> - [I also noticed an issue with the body text editor on card - will create a separate issue for it]

- Page Hero image field, and I liked the idea of keeping the URL option for now to not to break those websites that are using it.
- Pre-defined spacing on the top of headers.
- Default preview mode on blocks.
I also tested this branch on RTO website locally to make sure nothing is broken, everything seems to be working fine.

PR #160, opened 2021-03-25
ACF Button Block - make it unclickable in the editor side

zainabD:
I updated the theme.sass file, added `pointer-events: none;` to make the ACF button unclickable in the editor side. 

I figured that this is the cleanest and easiest way to make the button unclickable only on editor side.
I also tried to check if this is a post or page then remove the href in the php page, but that wouldn't work if we added more post types in the future.


wmcconnell (approved):
Works perfectly and is a really nice usability upgrade. Approved.

Issue #144, opened 2021-03-10
Need two lines (or an input type that allows for multiple lines) in a Hero title


zainabD:

wmcconnell:
Closing this one out. This feature was added to the Hero ACF group. Hero titles and sub-titles support the `<br>` tag now; sub-titles will automatically convert line breaks to `<br>` tags for manually breaking lines if needed.

PR #163, opened 2021-03-29
Multi-Button Heroes and Hero Image Alt Tags

wmcconnell:
Updates the Hero Image field group and associated code to allow for the multiple buttons we need for the SFIS (and other) degree pages. Also includes a small update to allow for `alt` tags on the hero image (now that we are using actual image tags, and not CSS background images).

- Update the Page Hero field group to include a new repeater field that repeats clones of the button field group, allowing users to enter an arbitrary amount of buttons in the hero area.
- Update the corresponding code to loop through that repeater field and render buttons. Added new code to handle the additional button values that the original fields did not include.
- Update the Page Hero field group to read the hero image as an array of data, and not just the URL to the image, so we can grab the `alt` or `description` values.
- Update the corresponding image tag code to include the `alt` value
- Maintain the original button fields for backwards compatibility (to be removed in a future update). The new hero will only fall back on the old button fields if there are no buttons in the new repeater field.

Changes **not** included here:

- Video heroes. Working on that one for a later update
- Full EWWW type support for `srcset` and WebP images for the hero. I believe we will need to replicate whatever EWWW does, since we are rendering our own `` tag here, and not using a block, or other WordPress technique, to insert the image on the page. That will also come in a future update once I take a look at the output of existing images.


zainabD (changes requested):
Can we remove the padding from the hero description: 
Screen Shot 2021-03-30 at 1 17 45 PM

- I noticed that the Single Word Highlight checking method is case sensitive, when typing the word in small letters we will get the error: 
`Notice: Undefined variable: title_string in /app/wp-content/themes/UDS-WordPress-Theme/templates-global/hero.php`
Can we add a message saying that this is case sensitive or maybe not accepting any values if don't match.
image

zainabD:
Screen Shot 2021-03-30 at 3 19 14 PM

zainabD:
This should also fix the issue #144

wmcconnell:
I've updated the Hero content area (title, text, buttons) to wrap it in a plain `container`, which seems to make it behave like an expected Bootstrap row/column.

zainabD (dismissed):

wmcconnell:
Added one more fix: Justin noticed that heroes without any buttons were having their titles moved down to the very bottom of the hero image. This was a result of a `margin-bottom: -2rem;` being applied to the content area. I updated our code to apply a particular class of `.has-buttons` when there are any buttons, and we target that more specific selector for the negative margin. I then add a margin to the bottom and left of the title when it's there by itself.

PR #164, opened 2021-03-29
Program's card

zainabD:
This is a new block for program's card, to allow uploading a hover gif image like the ones show on the new SFIS design.

- Registered a new ACF block called uds-program-card [**templates-blocks/program-card/**]
- Updated [**inc/uds-blocks.php**] to include the new block.
- Updated sass and css files to add the style of the new block.

https://user-images.githubusercontent.com/15880606/112912408-546aeb80-90ac-11eb-8b1b-58bac6f1e190.mov


wmcconnell (changes requested):
Zainab,

I pushed a commit to divide the ACF field into three tabs, to match some of our other blocks, and reduced the size of the image preview when editing. When you get a chance, sync the new ACF changes and tell me if you think that would work.

I was able to create cards, but did run into one issue: if the user forgets to fill in the 'Button Label and Link' field - leaving no link URL, target, etc. -  you get errors on the front end about missing array offsets:

Screen Shot 2021-03-29 at 5 48 20 PM

I don't know if this was inherited from my button code (and I'm going to go check my work on the Hero), but we should probably make sure those values are set before we try to render the button.

wmcconnell:
One more thing. Not a bug, just something I noticed. The `<p>` tag for the program name is limited to 50% of the width of the card. This was causing my title "Program three" to wrap, when "Program one" and "Program two" didn't. When I look at it, it seems like there's room to fit that title, but the 50% max width prevents it.

Screen Shot 2021-03-29 at 6 00 27 PM


Not sure if we wanted to do anything about this, but I didn't want to forget to mention it.

zainabD:
> One more thing. Not a bug, just something I noticed. The `<p>` tag for the program name is limited to 50% of the width of the card. This was causing my title "Program three" to wrap, when "Program one" and "Program two" didn't. When I look at it, it seems like there's room to fit that title, but the 50% max width prevents it.
> 
> Screen Shot 2021-03-29 at 6 00 27 PM
> 
> Not sure if we wanted to do anything about this, but I didn't want to forget to mention it.

Thank you Walt for your review. I added the 50% width just to follow the xd styling as you see they want us to break the title down at about 50% of the crad width. 
https://xd.adobe.com/view/1c93cf2b-d8f9-47f6-a2ba-07a396bf076c-49c0/ 
We can add an option for user to determine at what width should the title break if you think this is nessesary.

zainabD:
> Zainab,
> 
> I pushed a commit to divide the ACF field into three tabs, to match some of our other blocks, and reduced the size of the image preview when editing. When you get a chance, sync the new ACF changes and tell me if you think that would work.
> 
> I was able to create cards, but did run into one issue: if the user forgets to fill in the 'Button Label and Link' field - leaving no link URL, target, etc. - you get errors on the front end about missing array offsets:
> 
> Screen Shot 2021-03-29 at 5 48 20 PM
> 
> I don't know if this was inherited from my button code (and I'm going to go check my work on the Hero), but we should probably make sure those values are set before we try to render the button.

I saw this issue earlier when I was working but thought that this will never happen because why would the user leave the the button field empty :)
But you're right we should build clean code that gives no error in all situations. 
I will fix this.

PR #165, opened 2021-03-30
Add Named Anchors to Headings

wmcconnell:
Adds a new field to the Heading field group, and corresponding code in the template, to implement named anchors in our UDS Heading block. Users can enter an anchor name for a heading, and that name will be used as the heading's ID.

- Add new named anchor field to the field group
- Update the code to render an ID on headings if an anchor name was provided
- Changed the way we apply 'additional classes' to build an array, and then 'implode' it into a string once all additional classes have been evaluated.

### DIFF
--- a/acf-json/group_601dab5fa164e.json
+++ b/acf-json/group_601dab5fa164e.json
@@ -114,6 +114,25 @@
             "ui": 1,
             "ui_on_text": "",
             "ui_off_text": ""
+        },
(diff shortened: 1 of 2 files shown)


zainabD (dismissed):
Looks good to me, I only updated the offset and animation when goes to the target.

zainabD (approved):

PR #166, opened 2021-03-31
Fix anchor issues with tabs and accordion

zainabD:
I updated the jq code to fix the issue with anchor links on tabs and accordion. 
added the line to if statement: 
`(0 != target.offset().top)`


wmcconnell (approved):
Copied the default tab content from Storybook and compared the behavior of the tabs in one of my branches (without this fix), and this branch. Seems to solve the problem perfectly...no more scrolling to the top of the page. Approved so we can get that SFIS home page working!

PR #169, opened 2021-04-06
Remove margin form entry-footer section

zainabD:
I removed the mb-2 class from the container div in entry footer section to remove that extra margin that would appear when we're logged out of wp-admin. 

image


wmcconnell:
This work ended up in PR #170 as well, so I approved it there and am closing this one.

PR #170, opened 2021-04-06
override uds sass

zainabD:
I updated padding on UDS classes **.uds-hero-sm** and **.uds-list**


sass/theme/_theme.scss
```diff
@@ -112,3 +112,20 @@ position:relative;
 .acf-block-preview .card-button a{
 	pointer-events: none;
 }
+
+/*--------------------------------------------------------------
+Override UDS scss
+--------------------------------------------------------------*/
+
+//Override the UDS hero class on heroes.scss
+@media screen and (max-width: 576px){
```
wmcconnell, line 121: I think we should probably use the variables from `_variables.scss`, rather than actual pixel dimensions, on the small chance that someone changes the breakpoints. I think using `$uds-breakpoint-sm` will give us the same 576px, but also update itself if the small breakpoint is ever changed.

wmcconnell (changes requested):
I made a comment about the use of hard-coded pixel numbers in the `.scss` file. If we change that to use one of the variables, everything should be good.

wmcconnell (approved):
Updated the SCSS file to use variables, which was the change I had suggested. Noticed that this PR also includes the removal of some margin from the WordPress 'Edit' button. This duplicates work on another PR (#169). To make things a little smoother, I'm going to approve this one and close the other.

PR #172, opened 2021-04-07
External Link Updates

zainabD:
I updated the external link tags, created a separate switch for the target=_blank option. 
image


wmcconnell (changes requested):
It looks like our solution for hiding the external link controls on top-level items ends up preventing us from applying the icon to a sidebar menu that has only top-level items. I'm not sure if we should try to find a more specific class, or classes, or just solve this with some help text and documentation (or just allow the links anywhere and hope for the best)

wmcconnell:
Pushed updates to roll back our unsuccessful attempt to hide the external link controls in a very specific situation (top level items of only the main nav). Added ability to include the icon on top-level menu items, so that at least the controls are consistent (and on the off chance that someone asks us to do that very thing).

zainabD:
I can't find and easy and clean way to hide the external link options when CTA is on, so I think we can leave it as it is for now.

PR #173, opened 2021-04-09
Add Links to Hover Cards

### LINKED ISSUE #156: Add link to cards with hover action
In Cards block, if the hover option is checked, then buttons, tags and link fields will not be available.
we need to add a link field to link the card to somewhere when clicked on.

wmcconnell:
Resolves #156 

The hover card styling from the upstream Bootstrap theme handled the hover effect, but we needed to add the actual functionality to allow the entire card to be a link. Since 'hover' cards cannot have other interactive elements on them (buttons, links, etc.) my solution was to wrap the entire card content area in a `<a>` tag, and remove the usual link styling so that the cards look like they would if they were not a link.

- A new field in the ACF field group for supplying the target URL of the card. I'm using the ACF 'Link' field to allow for an actual URL, or for the user to select a page/post in the current site
- Updated markup and logic in the `card.php` template file to add the `<a>` tag around the content when the hover style has been selected
- New SCSS, in a specific `cards.scss` file, to **remove** the default link styles that happen when we add the `<a>` tag
- Updated `theme.scss` to add the hover card to our list of things with links that should not be followed on the admin side

Here's a video of them working:

https://user-images.githubusercontent.com/31013359/114243231-1342c900-9941-11eb-83c1-dff0d58e4f6b.mov


zainabD (approved):
Looks good to me, I tested this locally and it works with no issues.

PR #175, opened 2021-04-13
Tabs and Accordions block

### LINKED ISSUE #96: ACF Block: Accordion
We need to implement the foldable cards and accordions as shown in [Storybook here](https://unity.web.asu.edu/@asu-design-system/bootstrap4-theme/?path=/story/components-accordions--foldable-card). Since the content inside a foldable card can be arbitrary, we may want to try creating the wrapper and using `<InnerBlocks />` to handle the interior content.

zainabD:
I created a new block [**UDS Tabs and Accordions**] for tabs and accordion, the block include 5 different styles: 

1. Regular Maroon tabs
image

2. Video tabs (Like the ones used in SFIS)
image

3. Gold with dark background
image

4. Maroon with gray background
image

5. Accordion
image

- Created a new **acf-json** file for the set of fields used in this block.
- Registered a new ACF block [**templates-blocks/tabs/tabs.php**].
- Updated [**inc/uds-blocks.php**] to call the new registered block.
- Updated **Theme.scss** to add the new styles used in video tabs.
- Created a new js file to make the play button on video poster functioning [**js/play-button.js**].

Notes:
I added the ability of adding a pre-published post as a tab or accordion, just in case the tab's content has more complicated content that cannot be added through the wysiwyg editor:
image

In order to use the video tabs style, you have to have a pre-published post with a feature image uploaded to it.

Closes #96 when merged in.


zainabD:
This should also fix the issue #96

wmcconnell (changes requested):
I have a lot of items to mention, and I was going to get screen captures and videos to illustrate certain things. I think, however, it would just be easier to get together on Zoom and do a group review of this one. I'm going to put it on the agenda for our next developer meeting.

So, instead of a fancy review, I'll just put my initial notes here and I can demonstrate what I'm talking about when we meet:

**Overall**
- The tabs and accordions look good, and it's pretty neat to be able to switch between them. Nice work overall.
- There are a few interface and editing issues that, I think, prevent us from approving this as it is
- I encountered some issues when tabs and accordions are on the same page (odd scrolling, content in the tabs disappearing when an accordion is opened)
- I think a lot of what I encountered is probably due to items sharing IDs and confusing the javascript used to show/hide content. I need to investigate that further.

**Editor Interface**
- Tabs and accordions don't work as actual tabs and accordions in the editor
- The ACF interface needs some work, I think. If you use the visual mode where the inputs are in the sidebar, it's very difficult to use
- We should be able to collapse each row in the repeater so that just the title of the item is showing

**Misc**
- I don't think the gradient at the end of the tab bar is supposed to show unless there are more tabs hidden off-screen
- The gradient and left/right buttons are not showing up consistently on mobile size
- Choosing the gray or dark style didn't seem to have any effect at all. Would we solve this with our section block anyway?
- I think our scrolling code is in conflict with some of this, as opening tabs or accordion items sometimes caused my page to scroll to the wrong place (again, could be related to needing unique IDs on all items

**Philosophical Questions**
- Should we even have the video tabs here? I know they function like tabs, but they're really more of a stylized video playlist
- Should we go even further and break tabs and accordions into different blocks? Switching back and forth is really neat, but I wondered if the code for each one would be simpler, and more focused, if they were different blocks. It would also allow us to modify each of them separately if needed

Like I said, it would be easier to show these things. Let's plan on that for tomorrow.

zainabD:
- I updated the id of a single tab and added current time by microsecond to it, so it cannot be duplicated. 
- The tabs and accordion should work now in the editor. 
- Fixed the arrow issue, it will not show unless we have hidden items in the nav.
- It seems like the darkmode class and gray one were removed by the unity team and a bg-gray was used instead, it doesn't look right in the story book example. so I don't know if we still need to apply these styles. I removed these options from the block.
- I added all 4 colors for accordion. 
- We should be able to collapse each row in the repeater now.

wmcconnell (commented):
More notes!

I talked with Zainab about (most of) these, but we wanted to put them here so we didn't forget:

* 'Accordion' is misspelled in one place, but it's the place where we create unique IDs for all the tabs, so I figure we should fix it. It's on line 13 of the `templates-blocks/tabs.php` file.

Screen Shot 2021-08-17 at 3 05 39 PM


* Not all blocks that work in a post seem to work in a tab content area. When I made a two-column post with the Bootstrap Blocks, they didn't work inside a tab content area, and were just plain comments, like this (headings and paragraphs were made here, but not the container/row/columns)

Screen Shot 2021-08-17 at 2 18 25 PM

My columns still rendered, just not in columns:

Screen Shot 2021-08-17 at 3 04 19 PM


* Zainab said that the gradient/arrow control on tabs should not be appearing on desktop in the examples we looked at on my local site. It may be related to a message we saw in VS Code about the 'resize' method being deprecated in jQuery

* I had a general concern that using Posts for content results in actual posts that have URLs, and could be visited by a user - when they are meant only to be content within a tab. Would this technique work with 'Draft' posts as well as 'Private' ones? Could we build a tab block using ACF's `<InnerBlocks>` feature instead?

zainabD:
- I corrected the "Accordion" spelling. 
- Added "Edit" button to the pre-published post content. 
- Added filter "content to make all gutenberg block work as block not comments. 
- Fixed the gradient/arrow issue.

The <InnerBlocks> solution didn't work.

zainabD:
Per our last discussion I updated the following: 

- Removed the container/row/col wrappers from the regular tabs.
- Removed the video tab option, this will be built separately in another ACF block.
- Updated the content edit buttons: made it small and dark gray

zainabD:
I also added the following code so be can apply the bg-dark classon the tab container and get the right text color of the nav and content: 
```
.bg-dark{
	nav.uds-tabbed-panels{
		@extend .uds-tabbed-panels-dark;
	}
	>.tab-content{
		.tab-pane{
			@extend .text-white;
		}
	}
}
```
Screen Shot 2021-08-31 at 5 04 39 PM

zainabD:
I didn't add the tab styles as and option there but we can add any of the following classes to the container of the tabs to get the different tab styles:

- bg-dark
- bg-gray-1
- bg-gray-2

wmcconnell (dismissed):
Seems like we've addressed the things we were concerned about. Haven't had any trouble creating tabs or accordions on my end. I'd still like to see a way to avoid using posts for content, but this will allow us to build the things we need right now. Approved.

zainabD:
We built a separate block for foldable cards and another one for tabbed panels, no need to have this block anymore.

Issue #110, opened 2021-01-27
"Back to Top" Link


wmcconnell:
Please delete anything that does not apply for you!

## Issue is about..
  - [x ] a Feature Request

### Other Comments
We will need to implement the 'back to top' link [shown in the XD file](https://xd.adobe.com/view/56f6cb78-9af5-4b12-b4ce-ef319f71113f-03a5/screen/0fbfa336-9e35-485f-aad3-1c26219ce904/). Based on what the comments there say, the requirements would be:

- the control is hidden by default
- triggered by a scroll of just 1 pixel, but only when the user has already scrolled down 1.5 full screens (150% of the viewport height)
- disappears at some undocumented point. Probably when the vertical scroll falls below 150% again?

We should investigate any jQuery solutions, or other third-party libraries, or if this can be done entirely with CSS in modern browsers.

wmcconnell:
Closing, as this was merged a while ago. We do need to adjust it so that it only appears when the user has started scrolling _upwards_ after reaching 150% of the viewport height - but a new issue has already been created for that.

PR #176, opened 2021-04-13
Back to top button

zainabD:
I added a back to top button to header, gave it a fixed position, placed it on bottom right of window, 96px off the bottom as requested on the xd file: https://xd.adobe.com/view/56f6cb78-9af5-4b12-b4ce-ef319f71113f-03a5/screen/0fbfa336-9e35-485f-aad3-1c26219ce904/specs/
The button is hidden and will only show after scrolling through 150vh.

image


zainabD:
This should also fix the issue #110

js/back_to_top.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * JS for back to top button button.
+ *
+ */
+
+ ( function( $ ) {
+
+	$( window ).scroll(function() {
+
+		var vh = $( window ).height();
```
wmcconnell, line 12: If `var vh = $( window ).height();` is inside the function that fires on scroll, we're recalculating the window height on every scroll event - even though the window size is unlikely to change very often. I logged it to the console and it fired over a 100 times on a simple scroll down my sample page - but the height never changed.

Could we get away with putting the calculation outside the function that's firing on scroll?
It would mean that we only get the height on page load - and a change in window height would cause the 'back to top' link to appear at the wrong time (or not at all if you really shrink the window). I'm just wondering if that's better than running a function to calculate a value that's not going to change too much in regular use.

Or, split the difference and recalculate this only when the window size changes?

wmcconnell: By the way, everything works just fine and I'm happy to approve. I just noticed this and was wondering if we should change it.

wmcconnell:
During my review, I kept noticing that I was getting warnings about "scroll" and "resize" being deprecated - though I couldn't confirm this on the jQuery site. I was able to find this, however, on the jQuery Migrate site (emphasis mine):

> The .on() and .trigger() methods can set an event handler or generate an event for any event type, and should be used instead of the shortcut methods. This message also applies to the other event shorthands, including: blur, focus, focusin, focusout, **resize**, **scroll**, dblclick, mousedown, mouseup, mousemove, mouseover, mouseout, mouseenter, mouseleave, change, select, submit, keydown, keypress, keyup, and contextmenu.

I went ahead and changed the code to use the suggested `on("scroll", function())` and `on("resize", function())` syntax. The `resize` and `scroll` methods in jQuery are just shortcuts for the `on()` versions, as it turns out.

wmcconnell (approved):
Works great! Now that we're only recalculating the screen height when needed, I don't see any other issues. Approved.

PR #177, opened 2021-04-13
Fix addEventListener of null error on video hero

zainabD:
updated hero_video.js to only run when there is a video to avoid getting the error: 
`'Uncaught TypeError: Cannot read property 'addEventListener' of null'`


wmcconnell (approved):
This resolved the console error for me on pages with a static image hero, and did not seem to affect the video playback of my video hero. Approved.

Issue #178, opened 2021-04-14
Fix side scrolling caused by video on smaller breakpoints


zainabD:
The overflow: hidden is gone after certain point that is causing a side scroll issue on smaller breakpoints: 
![image](https://user-images.githubusercontent.com/15880606/114763706-f5e27600-9d17-11eb-9f5c-fdca1a2d7c22.png)

zainabD:
This issue was fixed.

PR #180, opened 2021-04-14
Fix the side scrolling on video hero

zainabD:
Updated the breakpoint when overflow switch to unset on hero-video to avoid the unwanted side scrolling on smaller breakpoints.
This should fix the issue #178


wmcconnell (approved):
Solves the problem and doesn't seem to cause any conflicts with standard image heroes. Approved.

Issue #134, opened 2021-02-26
Fix side scrolling issue caused by footer's .row divs


zainabD:
We temporarily fixed this by adding 
footer .row { margin-right:0; }
This is caused by one or more of the row divs in footer. 

![Screen Shot 2021-02-26 at 3 47 44 PM](https://user-images.githubusercontent.com/15880606/109363650-48d89a80-784a-11eb-8d74-272f78ad269d.png)

zainabD:
This issue is now fixed, I cannot recreate it anymore. 
I tested it out on RTO and devils drop off and it's no longer there.

zainabD:
Side scroll issue is back

zainabD:
The main cause of this is the padding: 0 added to last two footer in src/sass/extends/_globalfooter.scss: 
```
@media (max-width: 991.98px)
#footer-colophon {
    padding: 0;
}
 #footer-innovation{
    padding: 0;
}
```

PR #181, opened 2021-04-14
Fix the side-scroll issue caused by footer

zainabD:
This issue was caused by the padding: 0 value that was added to last two footer's container in src/sass/extends/_globalfooter.scss:

```
@media (max-width: 991.98px)
#footer-colophon {
    padding: 0;
}
 #footer-innovation{
    padding: 0;
}
```

I believe that this need to be updated from uds ends, but for now I added this to override the padding: 0 
```
@include media-breakpoint-down(md) {
  #footer-colophon, #footer-innovation{
		padding-right: 12px;
    padding-left: 12px;
  }
}
```

This should solve the issue #134


wmcconnell (approved):
Approving this right now as a matter of immediate need for the sites we're building. I'm a little concerned about 12px, as that's not really aligned with our 8-pixel grid (it's 0.75rem). I went looking through the UDS Bootstrap variables for a replacement for a hard-coded 12px, and discovered that they occasionally do use the 'magic number' of 0.75rem, so there's that.

Like I said, we need to fix this quickly - and it works together with the video resizing fix - so let's go ahead and do this for now.

zainabD:
The 12px was added by uds team before the padding:0 was added

PR #182, opened 2021-04-16
fixed issues with program cards

zainabD:
I updated the program cards css and js to apply the changes: 

- Made the gif image play once when hover, then replay when hover again (we need to use the non loop gif).
- Fixed the loading issue with program cards and the autoplay issue, as shown below: 

https://user-images.githubusercontent.com/15880606/114953726-dd548780-9e0d-11eb-8c67-64edf9ee7c3f.mov

- Updated button padding to avoid spacing issues when having longer words in title.
image

- Added the ability of adding a description to the card content so we can get a new cards style like the ones shown in SOS xd file:
image

- Updated the display name of the card to be uds overlay card (this should not break the existing content): 
image


wmcconnell (commented):
A couple of very small things:

- On the 'Images' tab, the label for 'image' is not capitalized. All other labels are, so we should probably change this one tp be capitalized as well.
- The original button icon field is named 'Button Icon', and we have the label prefix set to 'Button' for this cloned field group, so it says 'Button Button Icon'. Since we're using a tab to hold these fields, and it already says 'Button', we may not need to prefix the labels here.

I already have these changes made locally, so I can push them if we want.

zainabD:
> A couple of very small things:
> 
> * On the 'Images' tab, the label for 'image' is not capitalized. All other labels are, so we should probably change this one tp be capitalized as well.
> * The original button icon field is named 'Button Icon', and we have the label prefix set to 'Button' for this cloned field group, so it says 'Button Button Icon'. Since we're using a tab to hold these fields, and it already says 'Button', we may not need to prefix the labels here.
> 
> I already have these changes made locally, so I can push them if we want.

Yes please push them if you can. Thank you!

wmcconnell (approved):
Approved, as we do need these for our current site builds. I'm a little concerned, however, that they will end up being specific to KE/College of Global Futures. Once the dust settles on our site builds, we can evaluate again to see if there's anything we might need to update or enhance on these.

PR #183, opened 2021-04-16
Fix 'Phantom' Div Created by Banner Widget Area

wmcconnell:
When I created the widget for a site-wide banner, I included a checkbox for showing/hiding the banner - while still keeping the widget in the widget area. Normally, you would just remove a widget you no longer wanted, but I thought we might want to keep the widget there (in cases where you might have a seasonal banner that might come back every now and then), and activate it when needed.

Long story short, when you deactivated the widget it would not show - but it would sill render some of the wrapping HTML, including an empty container with a preset margin that would push your other content too far down the page. It hadn't come up on any of our existing sites, but it probably would have one day.

- Move the wrapping HTML out of a template partial, and into the widget class for the Banner; there is already code in the widget class to check and see if the checkbox is checked before rendering the widget. This means that none of the wrapping HTML is rendered when the widget is not active.


inc/class-uds-notification-banner.php
```diff
@@ -112,17 +112,25 @@ public function widget( $args, $instance ) {
...
-		</div>
-			<div class="banner-content">
-				<h3>$title</h3>
-				$body
-			</div>
-			$button_block
-			<div class="banner-close">
-				<button type="button" class="btn btn-circle btn-circle-alt-white close" aria-label="Close" onclick="event.target.parentNode.parentNode.style.display='none';">x</button>
+		<div class="container-fluid banner-$color">
+			<div class="container">
```
zainabD, line 116: I noticed that you added this **container-fluid** with the same background color as the the banner, so now we have a wider banner's background but the content and "x" button are still within the container width. 
Is this how banners should look like?

wmcconnell: I think so, based on the examples in the XD file. It even keeps the "nothing lines up at all" aesthetic of the original!

banners

inc/class-uds-notification-banner.php
```diff
@@ -112,17 +112,25 @@ public function widget( $args, $instance ) {
...
+				<div class="row">
+					<div class="col">
+						<div class="banner banner-$color" role="banner">
```
zainabD, line 119: Then I don't think we need this inner banner color if we're adding the color to the wrapper container

wmcconnell: Good point! That's been fixed, with the additional `banner-$color` removed from the actual banner `<div>`.

wmcconnell:
Went ahead and merged this, since it's a simple one that's effectively been reviewed and updated.

PR #185, opened 2021-04-22
Fix hero URL option

zainabD:
I noticed the following error when using URL option, since there is no **alt** added to the database: 
`Notice: Undefined index: alt in /app/wp-content/themes/UDS-WordPress-Theme/templates-global/hero.php `

To resolve this I added an if condition to only print that **alt** when there is an image uploaded through the media library.


wmcconnell (approved):
While we wait to find out if we should have blank `alt` tags for accessibility reasons, merging this to avoid any errors.

Issue #147, opened 2021-03-12
Modify Core Divider Block to use UDS Styles


wmcconnell:
We need to modify the core divider (aka `<hr />`) tag to support **only** our approved UDS styles:

1. Normal `<hr />` tag
2. Gold body copy divider (example found in Storybook)

wmcconnell:
This has been implemented as a block style and merged into develop. Closing.

PR #186, opened 2021-04-23
Updated core/separator block

zainabD:
In this pull request, I updated the native block(core/separator), removed the unwanted (**wide** and **dots**) styles, and added the uds gold style.
Per our last discussion (Walt and I), we decided that we don't need the other petern's dividers that show in storybook [morse-code-white, morse-code-black, topo-white, topo-black, network-white and network-black], as we can use the background section block instead.

image


zainabD:
This should fix the issue #147

sass/theme/_theme.scss
```diff
@@ -371,3 +371,18 @@ text-align: right;
...
+
+
+/*--------------------------------------------------------------
+Dividers
+--------------------------------------------------------------*/
+.is-style-copy-divider{
+	@extend .copy-divider;
+}
+.block-editor-block-list__layout .wp-block.is-style-copy-divider{
+	margin: 0;
```
wmcconnell, line 383: I think this line might be the issue with how the gold divider looks in the editor. Setting all margins to 0 removes the built-in 48px top/bottom that the default divider has. Could we get away with only setting `margin-right` and `margin-left` here?

wmcconnell (changes requested):
Everything seems to work, but the gold divider has no top/bottom margin in the Gutenberg editor (looks fine on the page), and sits right on top of whatever is below it (see screen shot). I think it's the result of removing the margins to prevent the existing `auto` margins from always centering things.


-- Screen shot --
Screen Shot 2021-04-23 at 4 16 01 PM

wmcconnell (approved):
Approving this with the caveat that we may need to go back and look at how we're enqueuing our block styles, as I think the style preview issue may be related to that.

PR #187, opened 2021-04-23
Fix color-picker.scss

zainabD:
We figured out in our wp meeting that the color picker classes where not generated correctly for a missing `$` before the variable [`name`], I added that missing `$`


wmcconnell (approved):
All the styles seem to be there now. Thanks, Zainab!

PR #189, opened 2021-04-26
Updated @asu-design-system

zainabD:
I ran npm install to pull the latest update form the unity design theme.


nathan-rollins:
I just noticed this pull request after I submitted my PR for the Cookie Consent. One minor correction I wanted to offer, in order to update the @asu-design-system packages to the latest versions that have been published, you actually need to run `npm update`. 

`npm install` will only reset the source code and copy the current code files from the installed @asu-design-system.

nathan-rollins:
Oh, actually it looks like I might be wrong about `npm install`. We have the packages pinned to "dev" and it looks like that allows these packages to update more often than normal. Regardless, we should always use `npm update` to update all our Javascript packages.

wmcconnell:
Closing this one as there is a more recent update in Nathan's branch from over the weekend. This is my fault, for letting this linger too long.

PR #191, opened 2021-04-28
News post type

zainabD:
Updates occured in this branch: 
- Removed the "hide page title" checkbox from "post" post type.
- Added "Author information" fields to "post" post type.
- Added "Story Hero" fields to "post" post type (Done by Steve)
- Added "Posts excerpt" field to "post" post type, and hid the WP Excerpt field, then added the following check: 
```
if ( ! $post->post_excerpt ) {
			$post_excerpt_acf = get_field( 'excerpt', $post->ID );
			$post->post_excerpt = $post_excerpt_acf;
		} else {
			update_field( 'excerpt', $post->post_excerpt, $post->ID );
			$post->post_excerpt = '';
		}
```
And that is to connect the ACF Excerpt field to WP Excerpt field. (please reach out for any question regarding this)

- Assigned the first image in the post to be the post featured image if that field was empty: 
```
 if ( ! has_post_thumbnail( $post->ID ) ) {
			if ( has_blocks( $post->post_content ) ) {
				$blocks = parse_blocks( $post->post_content );
				foreach ( $blocks as $value ) {
					if ( 'core/image' == $value['blockName'] ) {
								  $attached_image_id = $value['attrs']['id'];
								  break;
					}
				}
			}

			if ( $attached_image_id ) {
				   set_post_thumbnail( $post->ID, $attached_image_id );
			}
		}
```

-Updated Sass styling for single post (Done by Steve)
-Updated Single.php template (Merged Steve's work in there)


SteveRyan-ASU:
Adding some `the_excerpt` commentary: 

Zainab and I were talking about how to best deal with the requirement that the API payload from your news service should always require an excerpt to be created by the end-user.

The current branch tackles this problem in a direct way by presenting the end-user with a warning when the field is left blank. It prevents the post from being saved until this happens. (Not sure if an auto-save is triggered while the field is empty.)

Digging into the WordPress processes a little bit, we discovered the following:
See https://developer.wordpress.org/reference/functions/wp_trim_excerpt/ for context.

- Upon a template seeing `the_excerpt()`, if it were to encounter a blank in the database, WordPress would then call `wp_trim_excerpt` to automatically generate the 55-word auto-completed text from the body of the content window.
- It runs through a bunch of additional filters before delivering that payload as well. (Auto P, KSES, etc.)
- WP doesn't actually write the resulting string into the DB anyplace when it figures out that there's not an explicit string in the post_content column. 
- It instead relies on that complicated process to generate it on the fly each time it needs it.

So, my thoughts on how to solve this problem are these:

1. I **really** don't think we need to worry about the_excerpt field being blank in the API payload that's being sent out to the "spoke" news sites from the central hub.
2. If it's blank, WP will know what to do about it if the template is being rendered in the "normal" WP way.

Pausing here to let @nathan-rollins weigh in on the blank vs non-blank stuff first. :-)
 
Deep breath: If the output from the API is rending content outside of the normal WP processing methods and there is a direct need to have a value in the DB `post_excerpt` column at all times... I'm wondering if we could do something similar to the featured image treatment that's elsewhere in this branch.

- Rather than try to warn the user attempt to solve the problem programmatically.
- Upon a post save action and in the presence of a blank post excerpt field, maybe do the following:
- Replicate the process that WP uses to auto-generate the excerpt and force the output to be written into the DB. 

The one drawback with that approach would be that it would only happen once. So, if your opening words of the post were "Hello ASU" and you hit save, you'd get a post excerpt with the same string auto-generated (potentially). If you were then to go change that text to "Hello World" your excerpt would not automatically be updated to the new version. 

But that would solve the problem and 100% always have text in that field for your news API. And probably result in a better experience from an end-user perspective. (You'll wind up getting a lot of `.` excerpts if you throw up an error screen. 😄 )

SteveRyan-ASU:
Other more mundane stuff, checklist form.

- [ ] We should figure out what kind of interface we want for the social media icons at the top of the post. This could get tricky if the intent was to include SM "intent" actions that provide the user with an opportunity to share the content in their own social media. 
- [ ] Line 40 of `story-hero` should probably call `uds_wp_posted_on();` which produces the date along with the rest of the post meta information. (We can also consider altering that function to better define what that string should look like. It's currently pretty ugly.)
- [ ] My implementation of the story hero failed to do anything with the option to use a medium or large hero size with an image. An oversight that can be fixed by adding `.uds-story-hero-lg` to the top level of the hero when needed.
- [ ] The author meta that's output at the bottom of the post could use a little additional SASS to align the icons and the text. 
- [ ] If the post excerpt meta box persists, the title of the meta box should be singular. (Post excerpt as opposed to posts excerpt.)
- [ ] We should strip out the function in `/inc/setup` line 123 that includes a poorly formatted button as a filtered result for the excerpt. For now, that function could be left in place (pluggable in a child theme) but by default return empty. That should leave the excerpt otherwise unaltered.

SteveRyan-ASU:
Added a series of commits to this branch to accomplish the following.

- Remove the ACF post excerpt field from the theme entirely. After much discussion, there is no further need for it.
- Unregistered the deleted ACF field from the theme. No field = no endpoint.
- Reorganize some of the added code from this branch to better locations within the functions.php stack. 

More to follow from here. :-)

SteveRyan-ASU:
My last sequence of commits addresses the checklist of noted items above. There are a couple of things still left undone, but this is now committable in my view.
- We should rework the `uds_wp_posted_on()` function so that the article entry information on top of the headline is displayed in a format that better reflects the brand standard.
- If we choose NOT to display the category/tag information within this line, we need to align to WHERE this information is presented within the template. 
- The social media icons also need to be fully implemented in terms of a theme option panel or an individual setting on the post that allows the user to add the correct "intent to repost" code. This could also potentially be offloaded to a supported plugin that supports these activities already.

zainabD:
Thank you Steve for your work, here is a few things I noticed: 
- The ACF excerpt field is still there, I was wondering if that was a cause of the auto sync function that you added previously. 
- Yes I agree, we need to rework on the `uds_wp_posted_on()` function, that's why I removed it and printed the post date instead. I think it should print the post date and post status (for editor view), but not the post author.
- I already added the category information, and it is presented at the end of the single post. It is part of the `uds_wp_entry_footer()` function, I wrapped all categories with a card-tags div to give them the tags style: 
Screen Shot 2021-05-13 at 3 27 57 AM

nathan-rollins:
I'm reviewing this PR right now, but I wanted to go ahead and comment on the ACF Post Excerpt field. Even though it was removed from the theme's acf-json/ folder it appears to remain active in our site because it isn't removed automatically from the database. In fact, because I had edit and saved changes to three field groups on my local site, they all are listed in ACF as "My Fields" as well as "Active Fields". Very interesting. Looks like ACF tracks whether you directly edited or created field groups from your specific WP install.

inc/setup.php
```diff
@@ -118,23 +118,42 @@ function uds_wp_custom_excerpt_more( $more ) {
...
+if ( ! function_exists( 'uds_wp_custom_excerpt_length' ) ) {
 	/**
-	 * Adds a custom read more link to all excerpts, manually or automatically generated
+	 * Set a max number of words for excerpt.
 	 *
-	 * @param string $post_excerpt Posts's excerpt.
+	 * @param int $length the number of words in excerpt.
+	 */
+	function uds_wp_custom_excerpt_length( $length ) {
+		return 50;
```
nathan-rollins, line 130: 50 words is a bit long for our cards, that is around 300 characters. The XD design docs show News and Events cards running around 150 characters. How do you feel about trimming our excerpts to 30 words?

inc/wp-rest-api-extensions.php
```diff
@@ -0,0 +1,218 @@
...
+		array(
+			'get_callback' => 'uds_get_post_story_hero_api_field',
+			'schema' => null,
+		)
+	);
+}
+add_action( 'rest_api_init', 'uds_register_post_story_hero_api_fields' );
+
+/**
+ * Fetch and return the value of the uds_news_author ACF fields.
```
nathan-rollins, line 190: Oops! I forgot to update this function comment!

sass/blocks/_story-hero.scss
```diff
@@ -0,0 +1,83 @@
+/*--------------------------------------------------------------
```
nathan-rollins, line 1: Why do we have custom CSS here for our Story Hero? Is this because we haven't updated the UDS Bootstrap css yet?

single.php
```diff
@@ -23,30 +23,74 @@
...
+					echo '<p>' . $author_title . '</p>';
+				}
+				if ( $author_email || $author_phone ) {
+					echo '<p>';
+					if ( $author_email ) {
+						echo '<span class="fas fa-envelope-square"></span><a href="mailto:' . $author_email . '">' . $author_email . '</a>';
+					}
+					echo '</br>';
+					if ( $author_phone ) {
+						echo '<span class="fas fa-phone-square"></span><a href="tel:' . $author_phone . '">' . $author_phone . '</a>';
```
nathan-rollins, line 66: Note that the ACF field for author phone is a Number field (which is great by the way!) That means the phone value needs to be formatted here. The UDS standards specifies telephone numbers must use dashes, not parenthesis: xxx-xxx-xxxx

acf-json/group_6089d6189b18e.json
```diff
@@ -0,0 +1,103 @@
...
+            "placeholder": "",
+            "prepend": "",
+            "append": ""
+        },
+        {
+            "key": "field_6089d67c43514",
+            "label": "Phone",
+            "name": "uds_news_author_phone",
+            "type": "number",
+            "instructions": "",
```
nathan-rollins, line 66: Because this a Number field, the phone number must be entered as digits only (and correct formatting is then performed by the client). This field needs instructions explaining this. I was confused when first entering info. The field does not prevent typing the number as a string (including dashes), causing the field to not be saved.

templates-loop/content-card.php
```diff
@@ -0,0 +1,41 @@
...
+
+?>
+
+<div class="col col-12 col-md-6 col-lg-4 mb-4">
+
+	<div class="card card-story" id="post-<?php the_ID(); ?>" >
+
+		<?php
+		if ( has_post_thumbnail() ) {
+			echo '<img class="card-img-top" src="' . get_the_post_thumbnail_url( $post->ID, 'large' ) . '" alt="' . $alt_text . '"/>';
```
nathan-rollins, line 23: Is the large image preset the best preset to use for Card images? Given the size of the rendered image for Cards, medium is probably more efficient. We need to be conscientious about the file size we choose for rendered assets, like images. Choosing an excessively large image just wastes bandwidth, slowing down page load times and affecting our SEO and PageRank scores.

nathan-rollins (changes requested):
This code looks pretty good! Thank you both, @zainabD and @SteveRyan-ASU! I have noted some corrections and improvements that we may need.

nathan-rollins:
> Thank you Steve for your work, here is a few things I noticed:
> 
> * The ACF excerpt field is still there, I was wondering if that was a cause of the auto sync function that you added previously.

As I noted earlier, it appears the field is not removed automatically. I trashed and then permanently deleted the field from my local site.

> * Yes I agree, we need to rework on the `uds_wp_posted_on()` function, that's why I removed it and printed the post date instead. I think it should print the post date and post status (for editor view), but not the post author.

Hmm, maybe? I guess it might be handy for the Editor to see that. I presume you mean the status should not viewable to the public (all non-Editors).

> * I already added the category information, and it is presented at the end of the single post. It is part of the `uds_wp_entry_footer()` function, I wrapped all categories with a card-tags div to give them the tags style:

Nice.

That still leaves the question to how to handle tags. The ASU News content type includes several different taxonomies of tags/categories (campus, topic, etc.), but they are all then rendered together as a single set of tags on the Article. We do something similar by rendering the post tags with the categories.

inc/setup.php
SteveRyan-ASU: Totally works for me. I also felt like the 50-word mark was too big

[4.7 characters per word](https://www.google.com/search?q=average+word+length+english&oq=average+word+length+english&aqs=chrome..69i57j0i22i30l2j0i390l2.7815j0j7&sourceid=chrome&ie=UTF-8) x 30 words = 141 characters. Seems to be in the right sweet spot.

sass/blocks/_story-hero.scss
SteveRyan-ASU: Yes, that was the case when we started this pull request. It's been integrated into UDS-Bootstrap as of about a week ago, but I'm not 100% sure if we were able to update the NPM package in our theme to integrate it from the source.

If that happens, this SASS partial can be outright deleted. It's an exact duplicate of what was submitted to Bootstrap.

SteveRyan-ASU:
Making a "final pass" checklist based on the discussion above. 

- [ ] Trim returned excerpts to 30 words
- [ ] Update function comment for `uds_register_post_story_hero_api_field`
- [ ] Check to see if the story hero SASS has been included in the UDS-Bootstrap library. Use it if it's there.
- [ ] Declare and use an optimal image size for card images, or use the correct card image setting from the UDS Card block.
- [ ] Rework `uds_wp_posted_on()` to include the post date. If the user is logged in, also include the status of the post and the date of the last revision.
- [ ] Append the string of tags onto the end of the pill row of categories. Include an FA icon for a tag for some visual separation.

Also, look into dealing with the phone number meta box formatting. That might necessitate a "fast follower" PR that extends ACF to add a new field type. https://github.com/Log1x/acf-phone-number

sass/blocks/_story-hero.scss
SteveRyan-ASU: Double checked. The story hero SASS has not been included in the Bootstrap package that's a part of this PR. 

I'm fairly confident that when we merge #192 that it will be there. When that merge happens the correct thing to do would be to remove the `blocks/story-hero` partial completely.

inc/wp-rest-api-extensions.php
SteveRyan-ASU: Fixed. :-)

SteveRyan-ASU:
Adding a screenshot here of the work in progress. Visually, I think this will work for now. :-)
image

templates-loop/content-card.php
SteveRyan-ASU: Addressed. See comments below. :-)

SteveRyan-ASU:
Explaining some of what you see above:

- The `uds_posted_on` function was amended to display the last updated date only to logged in users.
- That same function was also amended to eliminate the "posted by" statement that linked back to the `author.php` template. (That's probably not a template that we want to promote the use of.)
- Added some icon language for the posted date and the tag row in the footer.
- Added some spacing around the author bio section

And some not-so-easily seen changes:

- Added a utility function to remove the hard-coded `width` and `height` attributes from the output of `get_the_post_thumbnail`. See `uds_wp_remove_thumbnail_height_width_attr` for details.
- That allowed for `templates-loop/content-card`. to produces a full and responsive srcset of images for the card thumbnails. The largest displayed image should still be the `medium_large` format.
- Those post thumbnails are now also lazy-loaded by default.

Other items cleaned up as requested above.

SteveRyan-ASU (approved):
Thumbs up from me. :-)

PR #192, opened 2021-05-01
Update UDS Bootstrap and install UDS Cookie Consent banner

nathan-rollins:
This PR fixes a couple build configuration errors and updates to the latest version of the UDS Bootstrap package (v3.11.0 currently).

* Font Awesome is no longer included in the bootstrap4-theme package. We now install it directly into the theme.
* Commit example browsersync options file
* Install and enable the UDS Cookie Consent banner
* Compile updated CSS and JS for the theme.


wmcconnell (approved):
Looks good to me:

- Checked the consent form on three browsers, in plain and private windows. The cookie consent appeared when it should have, and didn't appear when it shouldn't have.
- I also checked for any issues with existing FontAwesome icons on my local site, and there didn't seem to be any problems.
- After Nathan's update today, I was able to run `npm install` and `npm run build` without the errors from earlier this week

Knowing that we have a work-around for the build process, and that we need this code for the work being done on News, approving.

PR #193, opened 2021-05-03
Allow Customized Unit Name Text in Footer

wmcconnell:
Our current customizer and footer code simply repeats the site name beneath the logo in the ASU global footer. This was based on existing XD/Storybook examples, and is probably the 'correct' thing, but does not allow for the customization that our clients need. This PR updates both the Customizer and our footer code to allow for customized text in the footer.

**Changes proposed in this pull request:**
- Add two new settings for the Customizer:
  - `footer_unit_name_type`: for storing the choice between using the unit name, or custom text
  - `footer_unit_name_text`: for storing the custom text
- Add a new control in the Customizer to allow users to choose between repeating the site name, or entering customized text, for the footer text just below the logo
- Add a new control, hidden by default, for entering custom text. This field shows/hides itself based on the user's choice in the first field
- Updated logic in the footer file that checks the value of the first field and displays either the site name, or the custom text, based on the current setting

**Example Screenshot**:

Screen Shot 2021-05-03 at 3 20 47 PM

### DIFF
--- a/inc/customizer/customizer-settings.php
+++ b/inc/customizer/customizer-settings.php
@@ -525,6 +525,91 @@ if ( ! function_exists( 'uds_wp_register_theme_customizer_settings' ) ) {
 			)
 		);
 
+
+		/**
+		* Footer Alternate Unit Name Text
+		*/
+		$wp_customize->add_setting(
+			'footer_unit_name_type',
+			array(
+				'default'           => 'default',
+				'capability'        => 'edit_theme_options',
+				'type'              => 'theme_mod',
+				'sanitize_callback' => 'uds_wp_sanitize_nothing',
+				'transport'         => 'postMessage',
+			)
+		);
+
+		$wp_customize->add_control(
+			'footer_unit_name_type',
+			array(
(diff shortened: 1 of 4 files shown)


SteveRyan-ASU (approved):
Looks good Walt. A welcome addition for sure; I was contemplating offering to insert another action hook right here to replace this text with something else in a child theme. So, this saves a step.

Approving with the following commentary:

- Good to leave undocumented, but it's possible to alter the output of the string with inline HTML in the text field in the customizer. 
- For example, this does what you think it would in that space `<i style="font-weight:400;">Does this respect inline HTML?</i> Can we make <a href="#">a link</a> for example?`
- Wondering if it might be worth the effort in another PR to separate the controls for the logo/social media row into a different pane in the customizer. Maybe Footer: Logo/Social and Footer: Description/Links. (Each row gets its own panel that way.)

wmcconnell:
> Wondering if it might be worth the effort in another PR to separate the controls for the logo/social media row into a different pane in the customizer. Maybe Footer: Logo/Social and Footer: Description/Links. (Each row gets its own panel that way.)

I think you may be right about this one. I'll try it out on my local and see how it works. That section is getting a bit overstuffed at this point.

PR #194, opened 2021-05-04
Add support for posts

SteveRyan-ASU:
Starting this PR draft to allow for a little better visibility into the process of adding better support for `single.php` and the related views (archive pages). To date, I've committed code that does the following:

- Reworks the markup of `single.php` to accommodate a story hero `<section>` and separate `<article>` that encompasses the rest of the post.
- The post meta information (title, category/tags, author, etc) should all appear in the story hero element.
- The contents of `the_content()` will align with the story hero below it and will be surrounded by an article tag for accessibility purposes.

The story hero is an ACF meta box that will appear at the bottom of all posts. It provides options for:
- A "uds-approved" background-color choice or
- An uploaded image for the story hero

The story hero background image was intensionally created to be a separate image from whatever featured image is displayed for the post/page. This was done with the assumption that a featured image would typically not be of the same composition or art quality as a proper featured image. They would also not likely be composed with the same aspect ratio in mind. The user (in theory) would be free to use the same uploaded image in both places in the editor if they DID want the images to be one and the same. (This might be possible if the intended story hero size was "large.")

**Notes:**

- I spent some time making sure that the formatting of the page looked good in all device widths.
- The construction of the template intensionally omits any need for a Bootstrap block to define a container/row/column on the page. I didn't want our end users to have to worry about inserting those blocks when creating a news article.
- The page template itself also doesn't hard-code any bootstrap scaffolding into the design. 
   - This is so eventually we could consider supporting align-full and align-wide block widths as an added feature. 
   - This proved really difficult to pull off with the context of the story hero. So, I abandoned the effort for now.


SteveRyan-ASU:
Closing this request in favor of the combined group effort in #191.

PR #195, opened 2021-05-04
Fix PHP Warnings on Additional Classes

wmcconnell:
Some of our ACF blocks were throwing warnings about uninitialized variables with regard to retrieving the additional CSS classes from the `advanced` area of the Gutenberg editor. It turns out that I had copied/pasted the wrong bit of code to multiple blocks and left out the line that was initializing the variable used to hold those additional class names.

The following blocks were updated with this new code:

- Blockquote
- Button
- Card
- Content/Image Overlap
- Program (aka 'overlay') Cards

This was a code-only change, and **does not** require any re-syncing of  ACF field information. Changes proposed in this pull request:

- Update the code in all blocks that were not up to date already. The new code not only initializes the variable, but also includes sanitizing the input, and trimming off any extraneous whitespace.

### DIFF
--- a/templates-blocks/blockquote/blockquote.php
+++ b/templates-blocks/blockquote/blockquote.php
@@ -79,9 +79,10 @@ if ( $image ) {
 	}
 }
 
-// Retrieve additional classes from the 'advanced' field in the editor.
-if ( ! empty( $block['className'] ) ) {
-	$additional_classes = $block['className'];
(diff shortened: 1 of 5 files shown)


SteveRyan-ASU (approved):
Good remediation. I created a sample page with all of these blocks within. Query monitor immediately relayed the information that the notices are gone. 

Nice work. Quick and clean. :-)

PR #203, opened 2021-06-01
Pull updates from npm update

zainabD:


wmcconnell (approved):
Approved. The main change we want here is Zainab's fix for the Program Cards, since they are on the front page of all three sites we are currently building. We also brought in the latest upstream changes, which include brand-new styling rules for calendars, tables, in-page links, and more.

Issue #197, opened 2021-05-10
Fix Back-to-Top page button - incorrect scripting behavior


nathan-rollins:
FYI, we are not using the approved script for the Back-to-Top button provided in the UDS Bootstrap theme. Our button has missing logic in the script: the button should only appear after the user has scrolled past a specific range of the document AND they begin scrolling back up toward the top.

zainabD:
This was fixed in this PR: Back to top function #205

PR #205, opened 2021-06-03
Back to top function

zainabD:


zainabD:
This should fix the issue #197

nathan-rollins (approved):
Looks good. I just removed the obsolete code that was commented out. No need to keep that around, cluttering our code. We can always retrieve in the future from the git logs.

PR #207, opened 2021-06-07
Rename ACF Fields

nathan-rollins:
ACF documentation specifies field_names must not be reused in different field groups (with some exceptions.) But in general, field names must be unique, and we have been sloppy with our field names.

This PR attempts to fix this issues and possible future problems by renaming all fields to make them unique in the theme. These changes must also be applied to all field references (value accesses) in the theme templates.


nathan-rollins:
I have converted this PR to draft to show that it isn't ready for merger yet. (I should have done that from the start.)

nathan-rollins:
It seems this is really not an issue. I am closing this PR and consider the field names to be a non-issue.

Issue #133, opened 2021-02-26
Fix the space between navigation menu bar and hero section on mobile view


zainabD:
There is a space appears after 1260px breakpoint (between the nav menu and hero section), it was caused by a margin-top value that was given to <main> , the margin value changes at media (min-width: 992px) only. 
![Screen Shot 2021-02-26 at 3 38 49 PM](https://user-images.githubusercontent.com/15880606/109362706-d0bda500-7848-11eb-9d4e-5b04052c9a02.png)

zainabD:
This was resolved in this PR Header gap issue #208

PR #208, opened 2021-06-07
Header gap issue

zainabD:

- Updated header.php 

- Updated theme.scss to override the media query for breackpoint-down(lg) on src/sass/_global-header.scss
```
/*--------------------------------------------------------------
Override the desktop media query for header to fix the gap issue
--------------------------------------------------------------*/
// Desktop
@include media-breakpoint-up(xl) {
  #asu-header + div,
  #asu-header + main {
    @include transition;
    margin-top: 144px;
  }
}

@include media-breakpoint-down(lg) {
  #asu-header + div,
  #asu-header + main {
    @include transition;
    margin-top: 113px;
  }
}
```

- If we want to get this fix from the bootstrap theme, we can add a PR to update line 944 to 949 in src/sass/_global-header.scss
```
@include media-breakpoint-up(xl) {
  #asu-header + div,
  #asu-header + main {
    @include transition;
    margin-top: 144px;
  }
```

This should fix the issue #133


wmcconnell (approved):
Approving. Seems to fix the immediate issue of the gap above the hero image for both regular visitors, and those logged in and viewing the WP admin toolbar. Still need to have this fixed upstream, but this will solve our problem for now.

PR #210, opened 2021-06-07
Fix Auto Syncing Issues / Repair Page Hero JSON File

wmcconnell:
A fundamental misunderstanding of how ACF JSON sync was supposed to work led us to create a situation where we were automatically syncing changes in the wrong direction on our live websites. This led to automatic editing of data in our JSON files with out-of-date information in site databases. As it turns out, we should **never be syncing on a live website**, and there should **never be any field group, or field, definitions in a production website's database**.

- Remove the auto-syncing code that was actually meant for use on local development sites
- Update the JSON file for the Page Hero, which was a victim of one of these auto-sync issues and slipped past us.


SteveRyan-ASU (approved):
Approved in haste after our discussion. =)

nathan-rollins (approved):
Looks good to me. :)

PR #213, opened 2021-06-08
Fix alignment issue on desktop view

### LINKED ISSUE #75: ASU Header Logo and Content Area Not Aligned
Our content area is not left-aligning with the ASU logo:

Screen Shot 2020-12-02 at 9 50 10 AM

We need to determine if the problem is in the header markup, or the padding/margins from the body containers or rows and adjust accordingly.

zainabD:
This PR should fix the issue [ASU Header Logo and Content Area Not Aligned #75]
- I updated css files to make the header, footer and content all aligned in desktop and mobile view.
screencapture-localhost-3000-2021-06-09-15_46_43
Screen Shot 2021-06-08 at 5 13 24 PM

Closes #75 when merged in.


wmcconnell (approved):
Looks good and solves the current alignment issues. With things changing so much upstream, I'm sure we'll need to revisit this at some point after somebody "fixes" something else, but it does look much better with the various elements properly aligned -and it gets rid of one of our oldest issues!

PR #215, opened 2021-06-09
Initial creation of a UDS Grid Links block

### LINKED ISSUE #199: ACF Block: Grid Links
We should create a block that allows users to create the grid links [grid links shown here in Storybook](https://unity.web.asu.edu/@asu-design-system/bootstrap4-theme/?path=/docs/components-grid-links--grid-links-on-white-background).

To match the available styling, we should allow the user to select the following:

-  Background color of the grid area (white, light gray, dark gray)
- Text color (white or gold ). These may already be combined as part of the grid link styles, but that will need to be investigated

If possible, we might want to allow the user to configure these as well:

- Number of columns. Right now, all examples have two columns, but we should check the SCSS to see if other configurations are supported. If not, it may be tricky to build a flexible block
- Column/Row spacing. These values may be hard-coded as part of the current CSS, and it might be tricky to have the be user-adjustable without us resorting to computed, inline styles.

SteveRyan-ASU:
Closes #199 when merged.

This is an attempt to create a new UDS-Grid-Links block for our theme. 

## Options 

Included within the block are three possible data sources for the links

- terms from either the tag or category taxonomies
- plus a way to "roll your own" links using an ACF repeater field.

There are four possible column settings for the block along with the three possible text color settings (black text, gold text or white text). The block was specifically designed to sit well within a UDS Section Block to apply the recommended background colors from the Storybook examples. 

The only current way to deploy an icon alongside a link is to use the build your own method. (I didn't want to force a situation in which all terms in the default category or tag taxonomies were assigned a default icon via an ACF field.)

## Notes

I pulled the SASS partial for UDS-Grid-Links from UDS Bootstrap directly in order to include it directly within the theme. This was a deliberate move to avoid needing to update our theme to consume the file directly from the Bs4 library. (That would have probably added 50 additional "fixes" that needed to be addressed.) Like some of the other elements we now have in the theme, we can delete the SASS partial when we are ready to do the NPM pull. 

An additional note of caution: the clickable area for a grid link is the whole box for the link. So, when messing with Gutenberg settings and attempting to select the actual block for editing purposes, I frequently clicked a link and was taken to a new page in the browser. Not sure how to address that from a UI perspective.... :-)


src/sass/variables/_modals.scss
```diff
@@ -1 +1 @@
-$uds-modal-overlay-background-color: rgba(25, 25, 25, .75);
+$uds-modal-overlay-background-color: rgb(25, 25, 25, .75);
```
zainabD, line 1: I'm getting an error message in the compiler that this has to be **rgba** since it has 4 values not 3: 
 formatted: "Error: wrong number of arguments (4 for 3) for `rgb'\n" +
    '        on line 1 of src/sass/variables/_modals.scss\n' +
    '        from line 48 of src/sass/bootstrap-asu.scss\n' +
    '        from line 4 of sass/theme.scss\n' +
    '>> $uds-modal-overlay-background-color: rgb(25, 25, 25, .75);\n' +
    '\n' +

templates-blocks/grid-links/grid-links.php
```diff
@@ -0,0 +1,83 @@
...
+			$linkstring = '';
+
+			if ( ! empty( $icon ) ) {
+				$linkstring .= '<span class="fa-fw ' . $icon . '"></span>';
+			}
+
+			$link_url = $gridlink['url'];
+			$link_title = $gridlink['title'];
+			$link_target = $gridlink['target'] ? $gridlink['target'] : '_self';
+
```
zainabD, line 53: I noticed that when I'm on the preview mode, and trying to add a new row from the editor box on the right side, if the top row had an icon, and I try to add a new row, the block breaks, and it displays  the error: 
Screen Shot 2021-06-10 at 3 16 13 PM

Then in the page preview I see the following warnings for all 3 lines (50, 51 and 52):
Warning: Illegal string offset ‘url’ in /app/wp-content/themes/UDS-WordPress-Theme/templates-blocks/grid-links/grid-links.php on line 50 
Notice: Uninitialized string offset: 0 in /app/wp-content/themes/UDS-WordPress-Theme/templates-blocks/grid-links/grid-links.php on line 50

sass/blocks/_uds-grid-links.scss
```diff
@@ -0,0 +1,78 @@
...
+	.uds-grid-links.four-columns {
+		grid-template-columns: repeat(3, 1fr);
+	}
+}
+
+@include media-breakpoint-up(xl) {
+	.uds-grid-links.four-columns {
+		grid-template-columns: repeat(4, 1fr);
+	}
+}
```
zainabD, line 78: Try to add this line to avoid the issue on Gutenberg editor when trying to select the block, and accidentally click on the clickable area:
```
.acf-block-preview .uds-grid-links a {
	pointer-events: none;
}
```

templates-blocks/grid-links/grid-links.php
SteveRyan-ASU: This took me a bit to figure out. Within the `have_rows` loop in the PHP output, the variables for storing and building each grid link were not being reset with each iteration. As a result, there were intermittent errors that only occurred after the build of the initial link.

sass/blocks/_uds-grid-links.scss
SteveRyan-ASU: Great fix. Implemented and works well. :-)

SteveRyan-ASU:
Closing this PR in favor of a redone PR that implements this feature in a more clean way. (Doesn't involve an `npm build` in the process.) 

See PR #224 for additional details.

PR #216, opened 2021-06-11
Fix the featured image issue #214

### LINKED ISSUE #214: Automatic featured image function produces PHP notices if Yoast SEO isn't present within the install
## Issue is about..
  - [x] a Bug Report

When the Yoast SEO plugin is not present within a WordPress install, the following PHP notices are produced within the admin.
```
Undefined variable: wpseo_primary_term at wp-content/themes/UDS-WordPress-Theme/inc/template-tags.php:175
Trying to access array offset on value of type null  at wp-content/themes/UDS-WordPress-Theme/inc/template-tags.php:217
```

Also: there is now a dedicated function within the Yoast SEO plugin to check and retrieve the term marked as "primary." See the following Stack Exchange answer for context: https://stackoverflow.com/a/63072165/3431495

A fix for this should probably check for the existence of `yoast_get_primary_term_id()`, then if present, just call it directly to retrieve the term ID for the taxonomy that you are investigating. (Category, here.). From there, the ACF call to the page hero background image should still be possible as a "backup" featured image. 

In the event that Yoast SEO isn't being used, the right logic to follow would be to loop through ANY assigned category at that point for the existence of a hero image.

zainabD:
I updated the **uds_assign_featured_image** function to fix the issue mentioned in #214 when the Yost plugin is activated. 
If no primary category was selected, then the function will select the very first category in the list. 

Closes #214 when merged in.


wmcconnell (commented):
I installed the Query Monitor plugin on my local site, turned off Yoast, and was able to see the same errors shown on Steve's initial issue when I went to edit a post. I then switched my theme to Zainab's branch and saw a different error:

Screen Shot 2021-06-15 at 5 23 41 PM

Here's line 228 in context:

Screen Shot 2021-06-15 at 5 26 00 PM

Is anybody else getting this same thing?

zainabD:
> Query Monitor

I don't see that error you're seeing, although I deactivated Yeast and activated Query Monitor plugin.

zainabD:
I updated the **uds_assign_featured_image()** function added a condition to check `if ( $hero_asset_data )` exists before assign the ID value to featured image field of the post. 
This should correct the error mentioned by @wmcconnell

wmcconnell:
Closing this PR in favor of #225.

PR #217, opened 2021-06-15
Header text Highlights

### LINKED ISSUE #53: Custom Text Highlighting
We need the ability to add web standards highlighting (gold, black, etc.) to headings on our pages.

zainabD:
This PR should fix the issue [Custom Text Highlighting #53 ]

I updated the native heading block **(core/heading)**, added 3 highlight options (**Gold, Black and White**) , where you can select a specific word or sentence and highlight it.
This will wrap the selected word/words with a <span> tag and add the right class based on the selected color.

Screen Shot 2021-06-15 at 2 49 44 PM

Closes #53 when merged.


wmcconnell:
I updated this PR to add the 'active' state to our custom toolbar buttons. When the user selects text that already has one of our styles applied, the appropriate button(s) will have a different button state to show which styles have been applied to the selection, as shown here:

Screen Shot 2021-06-21 at 4 24 58 PM

It turns out that the default CSS styles for providing this feature, tied to the button's `isActive` property, require that you use SVG icons. It also turns out that the Dashicons used by WordPress are on their way out, as Gutenberg really wants you to use SVGs. I took the same Dashicon we were already using, and inserted an SVG version as a React element, and the feature worked as expected.

While I was there, I added some comments and cleaned up the indentation so it was consistent throughout.

wmcconnell (approved):
I've tried just about every combination of highlighting I can think of and everything works as it should. Approved. Nice work, Zainab!

PR #220, opened 2021-06-16
Core/image block

### LINKED ISSUE #219: Fix the native image block on ipad and mobile size
The core/image block sets a width and a height to the image which makes it shrinks when resizing the browser: 

Screen Shot 2021-06-16 at 4 09 05 PM

Screen Shot 2021-06-16 at 4 09 20 PM

Changing the image width and height from the block setting, or leaving the height empty should resolve the issue, but it breaks the block and give the option to recover it which will set it back to where it was. 
Screen Shot 2021-06-16 at 4 15 05 PM

Screen Shot 2021-06-16 at 4 14 32 PM

zainabD:
This PR should fix the issue #219 

I think the best solution for this issue for now is to set the image height to auto, and leave the width optional so the end used can resize the image as needed.
I added the following lines to sass file to override the image height in all core/image gutenberg blocks: 
```
.wp-block-image img{
	height: auto!important;
}
```

Closes #219 when merged in.


SteveRyan-ASU:
Pretty interesting discussion about why that CSS rule doesn't include `height:auto;` along with the Gutenberg default CSS rule of `.wp-block-image img {​max-width: 100%;}​` Apparently, there are a few edge cases in which you would want to make the images responsive with auto-width as opposed to auto-height. 

For the rabbit-hole, see: https://core.trac.wordpress.org/ticket/50909 :-) 

In any case, this quick fix is the intended solution, so I think we can/should proceed in this way. There are alternative ideas that would also work:
- Include the `img-resp` class from Bootstrap in the markup of the block.
- Use block filters and `preg_replace` to remove the attributes from the `` block.

... but I think this is the best. :-)

sass/theme/_theme.scss
```diff
@@ -451,3 +451,11 @@ footer #wrapper-footer-innovation .img-link{
 			padding-right: 0.15em;
 	}
 }
+
+
+/*--------------------------------------------------------------
+Always set the image height in core/image to auto
+--------------------------------------------------------------*/
+.wp-block-image img{
+	height: auto!important;
```
SteveRyan-ASU, line 460: Any particular reason for the `! important` tag here? 

There are some edge cases in which the responsive image settings would want to be enforced with `auto:width` instead of an `auto:height;` The important declaration would make it harder to override this elsewhere if desired.

SteveRyan-ASU (dismissed):
As it is, this solves the problem pretty quickly. :-)

wmcconnell (approved):
Re-approving after merging develop into this branch and reviewing it again. Steve had already approved this before, and I didn't see anything broken after getting the latest changes from develop.

PR #222, opened 2021-06-17
Fix Button/Headline Placement in Page Hero

wmcconnell:
For some reason (probably upstream changes), the placement of our headline and buttons in the page hero **on mobile** changed for the worse, pushing these items down into the content area like this:

Screen Shot 2021-06-17 at 2 36 23 PM

...and this:

Screen Shot 2021-06-17 at 2 41 13 PM

The placement of these items is affected by a combination of `margin-bottom` and/or `top` settings at various breakpoints. I went through each size of hero and checked to see which rules were causing trouble, and applied new SCSS to counter whatever was moving items down into the main content area. It looks good on all Hero sizes, with buttons or without:

Screen Shot 2021-06-17 at 2 35 05 PM

Screen Shot 2021-06-17 at 2 35 19 PM

This is not my favorite kind of pull request, as it's chasing after seemingly random changes - but, we have a ticket from a client who has noticed this, and we should get things looking better ASAP. Steve is working on some changes to the hero, so this fix may be short-lived, but we do need it now.

Changes propose:

- Updates to existing SCSS rules that determined the placement of the title and buttons in the page hero.


zainabD:
Screen Shot 2021-06-18 at 11 47 09 AM
Screen Shot 2021-06-18 at 11 49 29 AM
I don't think the margin of 2rem is a good solution for all button sizes, the large button doesn't seem to be in the middle of the hero bottom edge.

SteveRyan-ASU (dismissed):
These changes work to improve the situation with several sites that are in production.  Site by site remediation of the bug pointed out in this ticket may involve some combination of:
- the temporary reduction in the number of buttons present in each hero.
- reducing the size of all buttons so that they align properly. (Medium buttons instead of large ones)
- some site-specific CSS to address one-off hero styling in a specific site. (The customizer has a place for this CSS... great for applying a temporary bandage to a situation.)

Approving with an eye for refactoring the hero completely using markup now found within the Bootstrap 4 library, as of two days ago. (Reference: https://github.com/ASU/asu-unity-stack/pull/315)

SteveRyan-ASU (approved):
Still approved. ;-)

wmcconnell:
We've decided to go ahead with this now for the reasons that Steve outlined above. In summary, we have ways to fix certain cases on a site-by-site basis, and this should fix the most glaring issues we have on actual live sites right now. We'll be re-working the Hero section based on some upcoming upstream changes, so this is a temporary fix.

PR #223, opened 2021-06-21
Update bootstrap4-theme: from v3.20.8 to 4.2.10

SteveRyan-ASU:
Pull request to review and possibly commit a "naked" update of the upstream packages within NPM. What we are most concerned with updating here is the `bootstrap4-theme` package from ASU. 

Ran `npm update`, compiled the theme sass as a separate action, then ran `npm run build`.


src/sass/bootstrap-asu-extends.scss
```diff
@@ -20,7 +20,7 @@
 @import 'extends/heroes';
 @import 'extends/breadcrumb';
 @import 'extends/sidebar';
-@import 'extends/typography';
+// @import 'extends/typography';
```
SteveRyan-ASU, line 23: This partial was deleted completely from the SASS stack.

src/sass/bootstrap-asu-extends.scss
```diff
@@ -30,7 +30,7 @@
 @import 'extends/video';
 @import 'extends/card-and-image';
 @import 'extends/grid-links';
-@import 'extends/back-to-top-buttton';
+//@import 'extends/back-to-top-buttton';https://asudev.jira.com/browse/UDS-646
```
SteveRyan-ASU, line 33: This component was removed from Storybook as "not yet ready for prime time." The discussion within centered around the user experience of the button and what it was really attempting to accomplish. It also had significant accessibility issues (not reachable via the keyboard when it appeared on the screen.) 

Once these changes are merged, we will need to make a corresponding change to remove the JS that fires to trigger the Back to Top button from our theme.

src/sass/extends/_anchor-menu.scss
```diff
@@ -46,10 +46,19 @@
   .nav-link {
     color: $uds-color-base-gray-7;
     padding: $uds-size-spacing-3 0;
-    border-bottom: 1px solid $uds-color-base-gray-3;
     text-align: start;
     font-weight: 400;
 
+    &:not(:last-child) {
```
SteveRyan-ASU, line 52: - Feature not implemented in UDS-WordPress. Should be safe to proceed.
- Design of this feature raises questions in Storybook. (Should it really replace the header completely?) 

Recommend waiting to implement until further feedback obtained.

src/sass/extends/_blockquotes.scss
```diff
@@ -145,7 +149,11 @@ blockquote:before {
     img {
       max-width: 180px;
       height: auto;
+      background-repeat: no-repeat;
+      background-position: 50%;
       border-radius: 50%;
+      width: 180px;
```
SteveRyan-ASU, line 155: - https://github.com/ASU/asu-unity-stack/pull/296 for background.
- The "fix" here breaks the mobile display of block quotes by imposing a set height and width on the image. (Which should be responsive.)
- The intent is to allow non-square images to be included in the markup... but this falls short of that fix. 
- We have not implemented the new markup for Blockquotes in UDS-Wordpress as of yet, so we're still OK to merge this even though it's wrong.

src/sass/extends/_card-and-image.scss
```diff
@@ -1,10 +1,12 @@
 .uds-card-and-image {
   background-position: top;
+  background-size: cover;
```
SteveRyan-ASU, line 3: - Small fix for a "component" which should probably have never been implemented as a separate thing in the Bootstrap 4 library.
- OK to merge. It doesn't exist as a block within UDS-WordPress as of right now.

src/sass/extends/_cards.scss
```diff
@@ -642,6 +642,11 @@ Cards - Table of Contents
   .card.card-foldable {
     border-bottom: 1px solid $uds-color-base-gray-3;
   }
+  .card-header .card-icon {
```
SteveRyan-ASU, line 645: Small correction to icon cards. CSS only, shouldn't require a fix for the UDS-Card block.

src/sass/extends/_charts-and-graphs.scss
```diff
@@ -3,6 +3,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  max-width: 350px;
```
SteveRyan-ASU, line 6: - Small CSS fix, looks good.
- We don't have this implemented on our product as of now. OK to merge.

src/sass/extends/_grid-links.scss
```diff
@@ -54,56 +54,23 @@
...
-    @include two-col;
-  }
-
-  .uds-grid-links.mobile-at-lg.three-columns {
-    @include three-col;
-  }
-
-  .uds-grid-links.mobile-at-lg.four-columns {
-    @include four-col;
+  .uds-grid-links.three-columns,
```
SteveRyan-ASU, line 66: Contains further modification to the UDS Grid Links component to simplify what happens to the block when progressing down from desktop to mobile. 

OK to merge. UDS Grid Links block will take this into account when we merge it later.

src/sass/extends/_headings.scss
```diff
@@ -74,6 +73,14 @@ h5 {
   letter-spacing: $heading-five-letter-spacing;
 }
 
+h1 span[class^='highlight-'],
```
SteveRyan-ASU, line 76: - Multiple fixes since last update. See everything above May 10 here: https://github.com/ASU/asu-unity-stack/commits/dev/packages/bootstrap4-theme/src/scss/extends/_headings.scss
- Looks like they were solving padding & spacing issues when a highlighted header occurs after a paragraph. 
- Looks OK to merge. Markup for these changes not affected. 
- Worth noting that this work was done specifically within the context of highlights in a hero image.

src/sass/extends/_tooltips.scss
```diff
@@ -1,123 +1,121 @@
...
-  }
+div.uds-tooltip-container {
+  display: inline-block;
+  position: relative;
+}
 
-  // Display tooltip text on icon hover
-  &:hover &-text {
-    visibility: visible;
+button.uds-tooltip {
```
SteveRyan-ASU, line 6: - This is a relatively new element for UDS-Bootstrap.
- Looks like they needed to make multiple accessibility fixes to the initial creation. Looks solid now.
- The options for implementing a background color for the element rely on classes that are specifically written for this element. I'd like to figure out how we can instead use the background section block to do the same thing before we implement this as a block. 
- ... and that's if we decide to roll this out as a block. A tooltip is inherently an inline kind of element which may not behave or "fit" well within a block-centered design. 

OK to merge. We'll keep our eye on it.

src/sass/variables/_typography.scss
```diff
@@ -49,3 +49,12 @@ abbr {
   text-decoration-style: none !important;
   text-decoration-color: initial !important;
 }
+
+// Correct <strong> to weight 700. (BS4 default is 900.)
```
SteveRyan-ASU, line 53: This is fine and well documented in the SASS.

SteveRyan-ASU:
Actions required as a result of this merge:

- [ ] Need to remove the Back to Top button from our theme completely, or relocate the code in use to the theme instead of sourced from the Bootstrap 4 library. This update removes the story from the library completely.

Errors spotted:

- [ ] The fix for square images in the blockquote markup breaks the presentation on mobile. Will address prior to redesigning the UDS Blockquote block in our theme.

Questions raised:

1. I believe that the **anchor menu** would be a good candidate to include within the theme, as it solves some important secondary navigation issues. But the way that it is currently crafted raises some concerns which should probably be resolved before we build our block.
2. We should decide how (if?) we want to implement tooltips as an element within the theme.

SteveRyan-ASU:
Remove the back-to-top button by preventing the enqueue of the needed JS and removing the markup from `header.php` 

I purposefully left the script in place within the theme in case this decides to come back to life later: `wp-content/themes/UDS-WordPress-Theme/js/back_to_top.js`

src/sass/bootstrap-asu-extends.scss
wmcconnell: This makes sense. I see you've already create the PR to remove it ;)

src/sass/extends/_anchor-menu.scss
wmcconnell: Interestingly, I used one of these Anchor Menus in an attempt to improve our blog navigation - using it to display a list of categories as links across the top of `single.php`. It's not actually implemented, and I don't think my solution was the best, anyway. As you say, I don't think we're using this anywhere at the moment.

wmcconnell (approved):
Looks pretty tame to me. No changes to areas that have been problems in the past, such as the heroes, or the main header/nav area. Seems to be smaller CSS changes and, as Steve pointed out, the removal of the 'Back to Top' feature - which, while cool, is a bit intrusive and confusing.

Seems like we can merge this in and deal with the 'Back to Top' removal with Steve's PR.

PR #224, opened 2021-06-23
UDS Grid Links Block, Take 2

### LINKED ISSUE #199: ACF Block: Grid Links

SteveRyan-ASU:
Closes #199 when merged. Also invalidates the closed first attempt at this PR from #215. 

This implements the UDS Grid block in the theme and addresses all feedback provided by @zainabD within the first branch/attempt? Why multiple branches? We figured out that it would be a better strategy to incorporate upstream changes from UDS Bootstrap into the theme separately rather than integrate those changes + a new block in the same PR.

The upstream changes that contained the necessary code for this to work out of the box were added in PR #223.


wmcconnell (changes requested):
Looks good! I created all three types of grids: category, tags, and custom. Tried the various styles. No problems. I do have two small suggestions, however:

- We should set a `collapsed` field value for the repeater used when entering your own, custom links. That way, users can take advantage of the ability to collapse those entries down to one row each when the list is long. The Link field is the obvious candidate there.
- We should probably find a unique icon for this block. Find a good SVG icon at a site like [Iconify](https://iconify.design/icon-sets/), and then use [the instructions here](https://www.advancedcustomfields.com/resources/acf_register_block_type/).

zainabD (dismissed):
Looks good to me, nice work Steve!

SteveRyan-ASU:
New icon + collapsed field value.
image

wmcconnell:
So the issue I was describing earlier, with preview not working after you switch away from a set of custom links to tag/category, **only happens** when you have a value in the **icon** field for your custom link(s)

SteveRyan-ASU:
Noting the issue in play here. Steps to reproduce:

1. Creating a set of arbitrary links one of which includes an icon from Font Awesome. 
2. Toggle back to either the category or tag option in the block.

That should produce a "block can not be previewed" error like the one seen below.
image

Today, I discovered that this particular error is actually recoverable. 

1. Switch to the code editor. (Shift-Option-Command-M if you want the keyboard shortcut.)
2. Immediately switch back to the visual editor.

The block will be recovered and can be switched back to arbitrary link mode. The previously typed arbitrary options will also still be there.

zainabD:
There are two things I noticed that I missed the last time I reviewed this: 

- [ ] When adding an icon to **two lines** link, the icon should display to the left separately not as part of the text, just like how it looks in the xd file. We need to add the link text in a `<span>` tag with an `"inline-block" `display: 

Screen Shot 2021-07-28 at 3 26 23 PM

Here it is in the XD file: 
Screen Shot 2021-07-28 at 3 40 04 PM

- [ ] The other thing I noticed is that the links inside the box will always align vertically to top, even if there is more than one line link in the group of boxes in one row, is this how it suppose to work? or do we need to align them to middle.

Screen Shot 2021-07-28 at 3 22 56 PM

SteveRyan-ASU:
Good call @zainabD. Both of those problems would be best patched at the Bootstrap level - they exist within the Storybook implementation of the grid link element here as well. To see the effect, you may have to manipulate the HTML in the browser temporarily.

Link: https://unity.web.asu.edu/@asu-design-system/bootstrap4-theme/index.html?path=/story/components-grid-links--gold-text

The first of the two issues above has an easy fix, so I went ahead and patched it in the theme.

- Applying `display:flex` to the items within the anchor tag prevented the text from flowing under the icon.
- Because the icon has a height that is higher than the line height of the text, I added a little top margin to the SVG so it didn't look out of alignment.

I'd recommend that we keep everything top-aligned for simplicity. Deciding to do otherwise forces the element to need to be "content aware" and either align the icon to the top or to the middle, depending on whether the text is one line or two.

SteveRyan-ASU:
Merged in develop, resolved the problems noted above. :-)

PR #225, opened 2021-06-24
Rework uds_assign_featured_image() using native Yoast function.

### LINKED ISSUE #214: Automatic featured image function produces PHP notices if Yoast SEO isn't present within the install

SteveRyan-ASU:
Would close #214. 

This is an alternate approach presented in #216 which uses the native (and not very well documented) function within Yoast to return the featured image category.

I added checks to look for various states of the post creation including:

- when the post is new
- when the Yoast plugin isn't active
- when there is an active category but no image assigned to that category

This also prevents the function from being called during the `the_post` action hook, which fires in a whole bunch of places in the admin that aren't appropriate for this function. As it stands, this will only fire during the process of creating or editing a new post, upon a change of `post_status`.


SteveRyan-ASU:
I'll also add that the function contains a gap in the logic for assigning a featured image.

If there is no featured image, no found image blocks within the post and a single category assigned, the logic of the function won't fire to assign the category's assigned featured image. The reason for this is that the Yoast function for a featured_category won't be present when there is only a single category selected.

To correct this, we'd need to add a third clause to look for any assigned category and assuming that it finds one, grab the image.

SteveRyan-ASU:
Thanks for the discussion today @zainabD and @wmcconnell. This should be good to go and should be working according to the logic laid out in the function. Works safely with or without Yoast as an active plugin.

wmcconnell (approved):
Tested this on multiple posts, in various configurations and without Yoast active, and it is assigning the featured image as expected when I save the post. Met with Steve yesterday to go over a few things, and we didn't find any showstoppers. Approved.

PR #226, opened 2021-07-01
NPM install, rebuild SASS file, npm run build.

SteveRyan-ASU:
Rerunning NPM build processes to recompile theme source code.


wmcconnell (approved):
We need these fixes to repair damage that Walt did. Approved.

PR #230, opened 2021-07-14
contribute_button

zainabD:
I needed to override one of the theme functions in the child theme but was unable to do so for no `function_exists` checking has been done.
I added the line  `if ( ! function_exists( 'uds_wp_render_contribute_button' ) ) {` 
to check the function `uds_wp_render_contribute_button()`

### DIFF
--- a/inc/render-partials.php
+++ b/inc/render-partials.php
@@ -227,6 +227,7 @@ function uds_wp_render_footer_logo() {
  *
  * Takes the contribute_url setting and coditionally renders the button.
  */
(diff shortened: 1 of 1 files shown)


wmcconnell (approved):
Approved for emergency use on the TDN site. We'll need to visit this topic again, as there are multiple other similar functions in our theme.

SteveRyan-ASU:
Emergency use or not, these kinds of requests are no-brainer approvals for basically whomever needs them. Making things overwrite-able and hookable are always good ideas. :-)

PR #233, opened 2021-07-22
UDS Modals block

zainabD:
This PR contains the new ACF block for window modal.

Screen Shot 2021-07-22 at 3 00 36 PM

I added a new ACF block called "uds-modals" that contains a button and a Wysiwyg Editor field. 
This will allow us to add a button that opens a pop up window. 
It is possible to add more than one button in the same page, and each can open a different window.


SteveRyan-ASU:
With the update to `position:fixed` for the model window this can be merged. It works as expected. 

For future updates, I'd recommend an `<InnerBlocks>` approach as opposed to a WYSIWYG field for inputting the content. That would allow the user to select from a handful of blocks that make sense in a modal window generally. (Headline, paragraph, buttons, image). That would also keep the editing experience a little more consistent.

SteveRyan-ASU:
In playing around with the markup provided by Bootstrap, I also noticed that the modal window that is produced out of the box has a fixed height. The height is transformed well on a mobile display. But it leaves open lots of possibilities of odd looking displays for short or lengthy messages within the dialog.

image

Overriding the `height` and/or `max-width` values within `.uds-modal-container` in your child theme may prove valuable. Something like `.page-id-#### .uds-modal-container {height:auto;} should work for the specific page you are targeting.

SteveRyan-ASU (approved):
Looks good. Let's work on the `<InnerBlocks>` approach later for this one. :-)

PR #234, opened 2021-07-22
Blockquote char limit

zainabD:
I removed the limit from the quote text field.


SteveRyan-ASU:
Approved. No real reason for the initial quote text limit except some practical concerns about possibly using the block within another block.

SteveRyan-ASU (approved):

PR #239, opened 2021-08-02
Allow for More HTML Characters in Card Body

wmcconnell:
Switched from `wp_kses()` to `wp_kses_post()` to allow for more HTML characters/elements to appear in our cards. This arose when we needed to use certain tags, such as `<sup>` and `<sub>`, and HTML entities like `&trade;` and `&copy;` in certain cards.

The previous code was using `wp_kses()` without any allowed tags, which was a bit too restrictive for our needs. By switching to `wp_kses_post()`, we end up allowing the tags that WordPress has deemed acceptable for standard post content - which seems like a more reasonable stance to take.


zainabD (approved):
Looks Good!

PR #240, opened 2021-08-04
Allow More Special Characters and Formatting in Cards

wmcconnell:
We ran into a situation where my original, and quite restrictive, filtering of card titles and body text would not work with the client's actual content. In this particular case, we needed to allow subscript in both the title and body of a card in order to  display things like CO<sub>2</sub>.

My original code was based on the limited examples for card titles and body text, but it's become obvious that we need to allow some more flexibility here. The hope is that, by using the WYSIWYG editor for both titles and body text - and using WordPress's filter for common post content - we can safely handle just about any kind of text content that someone would want in a card,

**Changes requested in this PR:**

- Change the ACF field type for Card titles from `text` to `wysiwyg`, providing the 'Visual' and 'Text' tabs for entering more advanced markup in titles if needed. This was already done for the body text.
- Switch the function used to escape the title's output from `sanitize_text_field()` to `wp_kses_post()`, to prevent stripping out the markup we enabled by switching to the WYSIWYG editor in the first place
- Add a new toolbar option for the ACF WYSIWYG input to provide shortcuts to some of that advanced markup in the visual editing tab for both titles and body text. This new toolbar is called `Cards`, and provides buttons for:
  - bold
  - italic
  - superscript
  - subscript
  - special characters (through TinyMCE's modal dialog for easy selecting of special characters)

**Examples**

Screen Shot 2021-08-04 at 1 54 35 PM

A card with special characters and tags

---------

Screen Shot 2021-08-04 at 1 55 23 PM

The 'Cards' toolbar, and visual editing tab


wmcconnell:
Merged the latest develop (with the updated Bootstrap, etc.) into this feature branch. I didn't see any mention of this branch being behind develop, which I expected to see after updating develop yesterday. Is this related to new settings in our KE Org?

SteveRyan-ASU (approved):
Thumbs up from me. Nice work to keep the admin panel for the cards still very clean and organized. Does all of the things as advertised.

zainabD (approved):
Looks good!

PR #245, opened 2021-08-09
Update PHP sniffer installer for compatibility with Composer v2

SteveRyan-ASU:
Following a wipe of my machine, I had to reinstall composer at the global machine level. The version that was installed was >2.0, which broke the `composer check:cs` linting capability. 

Updating the `phpcodesniffer-composer-installer` to >0.7 and rerunning `composer update` restored those abilities.


zainabD (approved):
Works with no issues.

PR #246, opened 2021-08-09
Remove Support for Block Widget Editor

wmcconnell:
To **temporarily** prevent the odd redirects we've been encountering when attempting to open the widget screen, this simple update removes support for the new block-based widget editor, therefore falling back to the traditional widget editor.

Changes in this PR:

- Added code to remove support for the block-based editor in `inc\setup.php`
- removed that same existing code from `inc/widgets.php`, as it didn't seem to be working when placed there

This PR is just fix the fact that you can't add widgets at all in v5.8 of WordPress, and we should fix this more completely when time allows. I'm still unsure about what, exactly, is causing the problem. Google searching points towards TinyMCE (used in the only widget we provide), but removing that particular widget (or trying a simple text input) did not solve the problem.

### DIFF
--- a/inc/setup.php
+++ b/inc/setup.php
@@ -24,6 +24,12 @@ if ( ! function_exists( 'uds_wp_setup' ) ) {
 	 * as indicating support for post thumbnails.
 	 */
 	function uds_wp_setup() {
+
+		/**
+		 * Remove support for the Block-based widget editor introduced in WP v5.8
+		 */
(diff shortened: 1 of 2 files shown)


zainabD (approved):
Looks good, but I think we should add this to function.php so we don't forget about fixing it later. 
Or maybe create an issue in our queue for it.

SteveRyan-ASU:
Added issue in the queue to eventually fix this problem. See #250 for details.

PR #242, opened 2021-08-05
Revert "UDS-Hero metabox deprecation, replacement with blocks"

SteveRyan-ASU:
Reverts ASU-KE/UDS-WordPress-Theme#241. 

Pull request was not ready for deployment.

PR #235, opened 2021-07-27
Update bootstrap4-theme: from v4.2.10 to 6.3.3

SteveRyan-ASU:
Pull request to review the next update of the upstream UDS Bootstrap package from the Unity project.
Ran `npm install`, compiled the theme sass as a separate action, then ran `npm run build`.

PR #248, opened 2021-08-12
Reapply upboot changes to theme. (UDS-Bootstrap4 v6.3.3 targeted.)

SteveRyan-ASU:
We had reviewed and committed the work from this branch in #235. Then subsequently, we accidentally reverted that work in #242. This PR should recommit the code uplift back to the project. 

See PR #235 for the review of everything that should be happening here (again). 

:-/


wmcconnell (approved):
Returning our reverted changes into develop. Looks familiar, and does include the new Hero CSS we need to get Steve's Hero features working.

PR #251, opened 2021-08-13
CTA Buttons - main menu

zainabD:
In this PR, I activated the CTA option for first level menu items. 
- The option disappears when the first level item is a dropdown menu.
- All CTA buttons will show at the end of the menu, even if it was added to the middle of the menu.

Screen Shot 2021-08-13 at 10 38 24 AM

- On mobile view all CTA buttons show on one line at the end of the menu.
Screen Shot 2021-08-13 at 10 39 56 AM


wmcconnell (approved):
We still have some things to figure out when it comes to the CTA buttons, but this PR gets us the ones we are most likely to need - and the ones that people are actually waiting on. Was able to create all the required combinations of buttons on my dev site, and Zainab fixed the one error I encountered. Approved.

PR #252, opened 2021-08-17
Fix Content Placement in Large Heroes

wmcconnell:
After a recent update, Nic over at Cronkite said all his heroes were broken. The basic problem was caused by his custom Hero template in a child theme not being up to date with recent changes in our theme. Even after solving that issue, however, we discovered that the large size hero had a problem aligning its content. Heading, text, and buttons were effectively being centered, like this:

Screen Shot 2021-08-17 at 4 16 34 PM

After a working session with Nic and Steve Ryan, we determined that the `<div>` surrounding the content in Large heroes was set to a fixed width, instead of the `100%` used on other hero sizes. The `auto` left and right margins on the parent container were centering this `<div>` on the page.

- Change the fixed with `800px` of the content area with `100%`, which ended up with this:

Screen Shot 2021-08-17 at 4 18 12 PM

P.S. Nic has already tested my feature branch on one of his development servers, and it did solve his problem. I have not noticed any effects on other heroes on my local dev


SteveRyan-ASU (approved):
Such a small thing to have overlooked, but I believe this was the root of the problem. Approved.

PR #254, opened 2021-08-18
Repair 404 Page Layout

wmcconnell:
Our 404 page is essentially a blank page with a large UDS Hero. This means it was affected by the recent upstream changes and was no longer rendering correctly:

Screen Shot 2021-08-18 at 11 41 35 AM

This pull request repairs the current broken state of the 404, and applies some markup changes to make it more consistent with how we handle `page.php`. After applying changes, we have something that looks more respectable:

Screen Shot 2021-08-18 at 11 41 54 AM

**Note**: this fix does not deal with left-aligning the text and form. We already have #252 to deal with that specific issue.

Changes proposed in the PR:

- Add the `v1` hero classes where needed to move the content back to the desired location
- Wrap all the content in the same `<main>` tag used by our page template to return the left/right margins of `auto`, and prevent the hero image from sticking on the left at very large screen sizes
- Remove the `wrapper` div that previously held all the content. It was bringing in unwanted padding on the left/right at most breakpoints
- Clean up and re-indent the markup
- General cleanup of extraneous spaces in class names, and attributes that I didn't think were needed (mainly `title`)


SteveRyan-ASU (approved):
Looks like a good fix. Discussed with @wmcconnell in Slack. Approved.

PR #255, opened 2021-08-18
Restructure Archive.php to Fix Hero Placement

wmcconnell:
While fixing some issues with our 404 and hero templates, I noticed that heroes on our category (aka 'archive') page were not staying centered on the page at very wide screen widths, but were being stuck to the left side of the screen:

Screen Shot 2021-08-18 at 2 39 58 PM

After adjusting the markup (see below), I was able to get this to behave as I believe heroes are supposed to behave in 2.0 - maxing out at 1920px and staying effectively centered on screen, as shown here:

Screen Shot 2021-08-18 at 2 51 48 PM

Changes proposed in this PR include:

- move the current markup inside the existing `<main>` tag, as we do on other pages (`page.php`, `404.php`, etc.). It was outside that tag on this page.
- Remove the class "container" from that `<main>` tag. It was being applied as an argument to the `post_class()` method, but adding a container at that point added width constraints we don't want. Also, that particular argument was not included on those other pages.

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -13,18 +13,13 @@ defined( 'ABSPATH' ) || exit;
 get_header();
 
 ?>
-<?php get_template_part( 'templates-global/hero' ); ?>
-<main id="skip-to-content" <?php post_class( 'container' ); ?>>
 
-
+<main id="skip-to-content" <?php post_class(); ?>>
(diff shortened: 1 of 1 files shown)


zainabD (approved):
Good catch!

PR #256, opened 2021-08-19
Add Full-Width Option to Buttons

wmcconnell:
I had a request from Justin to allow for a column of equal-width buttons to match a design we were converting to the new standards. The width of a default UDS button is based entirely on the width of the button label (and icon, if present), so it's difficult to create two buttons of the same width.

I added a checkbox, which defaults to 'off', to apply Bootstrap's `btn-block` class on any UDS button, allowing for a button to take 100% of the width of its parent:

Screen Shot 2021-08-19 at 11 07 36 AM

 This allows for buttons like the one shown in the sidebar in this image:

Screen Shot 2021-08-19 at 11 03 02 AM

They stay full-width on mobile, resulting in this:

Screen Shot 2021-08-19 at 11 03 23 AM

As I mentioned above, the Link Grid is a better, and more compliant, method. I wouldn't use full-width buttons to create giant grids of buttons (although you could do it by creating multi-column layouts of full-width buttons). However, there may be some cases where we would like two buttons of the same width for aesthetic reasons.

Changes suggested in this PR:

- Add a Yes/No field to the UDS Button field group for enabling full-width buttons. Default value is 'No'
- Apply Bootstrap's `btn-block` class to any button where the full-width value has been set to 'Yes'

Notes:

- I'm using a Bootstrap class here because I'm not sure about our theme/block support for the WordPress classes for full-width items. I'm open to other methods, but if we're based on Bootstrap we might as well take advantage of the available classes.

### DIFF
--- a/acf-json/group_601db250e6c94.json
+++ b/acf-json/group_601db250e6c94.json
@@ -107,6 +107,25 @@
             "prepend": "",
             "append": "",
             "maxlength": ""
+        },
+        {
+            "key": "field_611d94a71a497",
+            "label": "Full Width",
+            "name": "full_width",
+            "type": "true_false",
+            "instructions": "Make button the full width of its parent container?",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
+            "message": "",
+            "default_value": 0,
+            "ui": 1,
+            "ui_on_text": "",
+            "ui_off_text": ""
         }
     ],
     "location": [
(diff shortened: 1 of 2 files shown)


zainabD (approved):
Looks good!
Also if we want to use the bootstrap classes instead, we can move the line `<?php echo $additional_classes; ?>` to the **<a> tag** level, then add the **w-100** class to the additional classes field, that should do the same. 
But I think the switch button is much neat and user friendly solution.

wmcconnell:
While looking this over one last time before planning to approve it, I realized the following:

- This doesn't work on Cards, where the card button classes are going to prevent the full-width class from working. This would result in an inconsistent experience for our users
- We might need to modify other button rendering classes, such as the Hero, to accommodate this idea - making it a larger change than I wanted

I'm going to close this one in favor of working on a 'button group' block where the alignment and size of the buttons could be controlled by that block, and not by modifying the base button class.

PR #257, opened 2021-08-19
New Block: UDS Banner

wmcconnell:
While we have a widget, and a widget area, for banners across multiple pages - we don't have a way to easily insert a single UDS banner on a page. With the COVID banner seeming demoted to more of an 'alert' status, but still retaining banner styles, I figured we should also have a block for adding banners.

Already wrapped in a fluid container, the block can be used directly at the root level of a page for a max-width banner, like this:

Screen Shot 2021-08-19 at 4 30 39 PM

Or within a container/row/column structure to fit within the content area, like this:

Screen Shot 2021-08-19 at 4 33 01 PM

Changes proposed in this PR:

- Add a new block version of the existing banner widget, with the following features:
  - All features of the widget (multiple banner colors, icons, up to two buttons)
  - Ability to hid the 'Close' button to make banners that cannot be dismissed (i.e. the COVID banner)

### DIFF
--- a/acf-json/group_611ec0462e2e5.json
+++ b/acf-json/group_611ec0462e2e5.json
@@ -1,0 +1,281 @@
+{
+    "key": "group_611ec0462e2e5",
+    "title": "UDS: Banner Block",
+    "fields": [
+        {
+            "key": "field_611ec04631bc0",
+            "label": "Color",
+            "name": "uds_banner_color",
+            "type": "select",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
+                "id": ""
+            },
+            "choices": {
+                "green": "Green",
+                "orange": "Orange",
+                "blue": "Blue",
+                "gray": "Gray",
+                "black": "Black"
+            },
+            "default_value": "green",
+            "allow_null": 0,
+            "multiple": 0,
+            "ui": 1,
+            "ajax": 0,
+            "return_format": "value",
+            "placeholder": ""
+        },
+        {
+            "key": "field_611ec04631bc8",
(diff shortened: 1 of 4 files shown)


zainabD (changes requested):
It looks good! 
I just noticed that the content of the banner is not aligned with the page content, since we're using **container** inside a **container-fluid** for the banner, so it adds padding-x twice. 
I would add a **px-0** to one of the containers to fix the issue: 
Screen Shot 2021-08-20 at 4 38 16 PM


After adding **px-0** to the container wrapper: 
Screen Shot 2021-08-20 at 4 37 41 PM

wmcconnell:
Per Zainab's note, I added the `px-0` class to the inner container. This makes the full-width version align the content with the logo, like this:

Screen Shot 2021-08-23 at 3 16 45 PM

While a banner inside the content area aligns like this:

Screen Shot 2021-08-23 at 3 17 03 PM

I think that second one is what we would want inside the content area. It would look strange if we aligned the content in that case - as we'd either have no padding inside the banner, or the banner color would "leak" out of the content area.

zainabD (approved):
Looks good!

wmcconnell:
Forgot that I had some ACF field group changes on local that I needed to finish before asking for a review, so I'm re-re-requesting a review.

- Added tabs to split up style and content fields
- Added help text
- Updated the field group description
- Switched to WYSIWYG editor for the title, just in case they need special characters in the future

wmcconnell:
I added default text of 'Button One' and 'Button Two' for the buttons, so they will be created as full-sized buttons with labels, like this...

Screen Shot 2021-08-25 at 9 30 35 AM

zainabD (approved):
All look good!

PR #259, opened 2021-08-26
Background section block

zainabD:
In this PR I updated the **UDS Background Section Block** to make it align better with the rest of the content and to fix a side-scrolling issue that was caused by this block in a certain situation. 

Fixes that appears in this PR: 
The content of the container div in this block did not align well with the rest of the page content, and it varies from one block to another depends on whether we're using a background color for the content or not:
screencapture-new-wprig-lndo-site-test-page-2021-08-25-13_34_54

To fix this issue I updated the template of the `<innerblock>` in **background-section.php**, added a **core/group** block inside the container, and assigned all the padding and background colors to the inner **group** div instead of the **container** div.

screencapture-new-wprig-lndo-site-test-page-2021-08-26-11_30_56

Another issue appeared with this block is a side-scrolling on 1260px screen width, and this only appeared when adding a rows bootstrap block inside the container div: 
Screen Shot 2021-08-25 at 1 22 18 PM

I added a `overflow: hidden` to the **.uds-section** class, but then I noticed that the **core/group** block that I added earlier fixed that issue. 
Anyway, I think it is better to leave the `overflow:hidden` there as a fix of the existing background blocks that has rows in there, because after merging this PR in, the existing blocks in our current websites will not get the new template updates, but the new created blocks will.


wmcconnell (commented):
I noticed one thing right as I was ready to approve this one: on blocks that are already made, and are using one of the UDS textures as a background, the color behind the text will disappear when you switch to this branch:

Screen Shot 2021-10-04 at 5 33 20 PM

However, the white area is still showing in the editor:

Screen Shot 2021-10-04 at 5 33 03 PM

We can probably fix whatever is causing the block to look different in the editor, but I'm worried about introducing a block that is not backwards compatible. If someone has used one of the background textures in their block, updating to this version of the block would change how their pages look...although I don't know if anyone is using this block regularly, so now might be the time to change it - before there are more of them out there.

PR #261, opened 2021-08-27
Header menu when logged in

zainabD:
In this PR I updated the header menu position in mobile view when logged in to WP dashboard. 
Currently we have this gap before the menu if the user was logged into the wordpress dashboard, it only appears on mobile view when you scroll down. 
Screen Shot 2021-08-27 at 8 44 29 AM


wmcconnell (approved):
Looks good! Approved during weekly dev meeting.

PR #262, opened 2021-08-31
Single news page

### LINKED ISSUE #229: Single Post Template Doesn't Handle Floated Images Well
## Issue is about..
  - [ x  ] a Bug Report
  - [x] a Feature Request
  - [ x ] a Suggestion
  - [ ] a Question
  - [ ] I need help

Our template for single posts, `single.php`, does not seem to handle floated images very well. In the contrived example post below (based on an actual news story), even rather small images can break out of the content area when there is not much text. In this example, I have even **reduced** the image size to make it look better!

Screen Shot 2021-07-02 at 3 13 37 PM

I believe we can fix this by applying the Bootstrap `clearfix` class to the wrapping `<article>` tag that holds the content. This approach worked in my testing, resulting in this, even with the images scaled back up to 100% of their original size:

Screen Shot 2021-07-02 at 3 18 29 PM

zainabD:
I updated the **single.php** page, wrapped the `<article>` into a container/row/col wrappers to avoid having some of the issues that caused by:
```
 .single article>* {
    max-width: 995px;
    margin-left: auto;
    margin-right: auto;
}
```

Issues I found in the current version of single.php: 
1- Blocks that had `d-inline-block `class don't stay within the 995px space of the page content
Screen Shot 2021-08-27 at 11 15 04 AM

2- The golden `<hr>` should not take the full width of the page
Screen Shot 2021-08-27 at 11 15 23 AM

3- If image was wrapped with `<picture>` tag, which is automatically added by the Eww image plugin, the image appears on the left side of the screen:
Screen Shot 2021-08-27 at 11 18 37 AM

4- Images with float doesn't work as expected: 
Screen Shot 2021-08-31 at 1 22 09 PM

One of the issues we discussed before about using a **container** wrapper in the single is the need of using the **container-fluid** within the page content, and I think that is still possible if we create a new ACF block that can only be used in the single page and should start with the end tags of container/row/col wrappers, something like this: 
```
</div></div></div>
<div class="container-fluid">
<InnerBlock/>
</div>
<div class="container single-news-post-container pt-md-4">
<div class="row">
<div class="col">
```

This PR should close #229 when merged in.


wmcconnell:
Pushed a commit that fixes the missing author information. The field names had been changed for the Infonet news service, but we had not updated our code to read those new names. Also added help text to the phone number field, since you must only enter numbers there if you want the `tel:` link to appear at all.

wmcconnell (approved):
I was able to recreate the issues listed in the PR, and have them resolved by the changes here. I did notice that `single.php` was checking to see if we had set the page title, but that field group only appears on pages, not the posts that would be displayed by `single.php`. I removed that code, as it's always going to evaluate to `false`.

I viewed, created, edited several posts without any obvious issues. Looking in to this also exposed an issue with category pages, so that's an added bonus. Approved.

PR #264, opened 2021-09-01
Grid links with bg image

zainabD:
This branch is based on Steve's branch **sryan-uds-grid-links-take2** #224 
I added an image field to the "create your own links" option, so we can add a background image to each link individually.

Screen Shot 2021-09-01 at 4 40 47 PM


wmcconnell:
I've added a fix for an issue I discovered: when using icons, the spacing was wrong due to the `<p>` tag that we added to support the image background.

Screen Shot 2021-09-09 at 5 25 45 PM

I updated the logic to remove the `<p>` tag whenever there was **not** an image background, so that we have the original markup. When there is an image background, we add the `<p>` tag so our image links look correct:

Screen Shot 2021-09-09 at 5 28 08 PM

wmcconnell (approved):
Reviewed personally, in with the group. I updated some of the logic myself, and merged in the latest develop. No issues creating grid links. Approving with an acknowledgement that I believe the image links should be their own block one day ('Image Grid') with options for various sizes and text placement, etc.

PR #265, opened 2021-09-02
Media Gallery block

zainabD:
I created a new ACF block for media gallery.
This block can be used to create a UDS image gallery without content: 
Screen Shot 2021-09-02 at 1 11 00 PM


wmcconnell (commented):
Looks pretty good. There are two things I wanted to point out:

### Buttons
I changed the markup to use a FontAwesome button "stack" to get the previous/next buttons that looked closer to the XD. We can always abandon this approach, but I wanted to see if I could get it to work. It looks like this:

Screen Shot 2021-10-05 at 4 53 30 PM

Should we continue with this method, or do we have a better way already implemented?

<hr />

### Thumbnails
The image thumbnails shrink on mobile screen, but the XD shows them always staying the same size on desktop or mobile - with fewer of them in mobile.

Screen Shot 2021-10-05 at 4 53 55 PM

I think they want the thumbnails to slide along just like the big images. However, I don't know when they would slide, or whether or not it would be one at a time or a whole "page" of them. We could check the Unity version, I supposed.

PR #268, opened 2021-09-08
gulfile updates

zainabD:
I updated the gulfile.js to make the gulp watch or watch-bs updates the needed files only without having to pull any files from upstream.


wmcconnell:
Just to have it on the record, here is a Google Drawing (available in our team drive) of the current build commands:

UDS Theme Gulp Process

PR #269, opened 2021-09-08
Fix Breaking Upstream Changes

wmcconnell:
With the recent bump to 0.25, some destructive upstream Bootstrap changes snuck in while we were distracted with making sure that the changes to the Hero markup didn't break everything. I've made some updates to remediate the effects of these upstream changes.

This PR proposes:

- Updates to the sidebar menu walker to apply the updated markup to our sidebar menus. The upstream sidebar was changed so that every single `<li>` in a menu now needs to be inside of a `<div class="nav-link-container">`, along with other changes. Our current CSS is not targeting these new items, resulting in generic (but functional!) menus like this:

Screen Shot 2021-09-08 at 10 51 28 AM

after updating the walker, the menus are restored to the proper 2.0 look:

Screen Shot 2021-09-08 at 10 51 06 AM

- Adding/Removing/Editing of the menu markup and CSS to fit the markup now found in storybook. Most of this had to do with how sub-menus are wrapped in collapsible cards

- The new menu structure also removes the menu title from within the menu markup. The new solution is to use a basic `<h3>` tag above the menu. Removed the markup and logic that would build a menu title if one was provided in the short-code
- Also added a new option in the short-code for adding a specific ID to a sidebar menu, should we ever need it.

- Fixes for button link colors within certain background classes. Someone updated background color classes (`bg-white`, `bg-gray-2`, etc.) to force specific colors for `<a>` tags. Because our button block uses styled `<a>` tags to make buttons, this was overriding our button link colors - leading to maroon text on maroon buttons, and other non-standard combinations:

Screen Shot 2021-09-08 at 11 14 22 AM

-I have applied very specific rules to those links, such as `a .btn.btn-maroon { color: white; }` so that our buttons will consistently have the proper text color regardless of their parent container(s) classes:

Screen Shot 2021-09-08 at 11 17 37 AM


wmcconnell:
One more gripe/fix:

- The sidebar CSS was changed so that the currently active item went from `.active` to `.is-active`. I wrote a filter to search/replace this value **only** in a sidebar menu so that the gold underline. I put that filter in  `inc\hooks.php`, as it seemed like a good place. If they change back to `.active`, we can simply delete that callback.

inc/hooks.php
```diff
@@ -49,3 +49,24 @@ function uds_wp_add_site_info() {
...
+				if ( 'active' === trim($val) ) {
+					$classes[$key] = 'is-active';
+				}
+			}
+		}
+
+		return $classes;
+	}
+}
+add_filter( 'nav_menu_css_class', 'uds_wp_replace_active', 20, 4);
```
zainabD, line 72: I don't think we need this function here in hooks.php, since we have created something similar on inc/navigation-widget.php
In **inc/navigation-widget.php**, you can change the line 44 to **is-active** and that should do the work.

```
/* Check if function exists. */
if ( ! function_exists( 'uds_wp_theme_special_nav_class' ) ) {
	/**
	 * Replace current-menu-item class with active class.
	 *
	 * @param classes $classes is an array of the CSS classes that are applied to the menu item's individual element.
	 *
	 * @param item    $item is the current menu item.
	 */
	function uds_wp_theme_special_nav_class( $classes, $item ) {
		if ( in_array( 'current-menu-item', $classes ) ) {
			$classes = array_diff( $classes, array( 'current-menu-item', 'active' ) );
			

                   $classes[] = 'active ';   <============ HERE - Change this to is-active



		} else if ( in_array( 'current_page_item', $classes ) ) {
			$classes = array_diff( $classes, array( 'current_page_item', 'active' ) );
			$classes[] = 'active ';
		}
		return $classes;
	}
}
```

inc/shortcodes.php
```diff
@@ -18,17 +18,20 @@ function uds_wordpress_shortcode_sidebar_menu( $atts ) {
 		array(
 			'menu'            => '',
 			'title'           => '',
+			'id'              => 'sidebar-menu',
 		),
```
zainabD, line 22: The id cannot be static, just in case if we have two sidebars in one page.

zainabD (changes requested):

inc/shortcodes.php
wmcconnell: Updated this with the following:

- Removed my custom filter for the 'is-active' class, and modified Zainab's existing code
- Removed the 'id' argument from the short code and used the menu's name as the ID, since the name is already available and is a perfect place to craft a custom name for a menu!

zainabD (changes requested):

inc/navigation-widget.php
```diff
@@ -41,10 +41,10 @@ function uds_wp_theme_custom_widget_title_tag( $args ) {
 	function uds_wp_theme_special_nav_class( $classes, $item ) {
 		if ( in_array( 'current-menu-item', $classes ) ) {
 			$classes = array_diff( $classes, array( 'current-menu-item', 'active' ) );
-			$classes[] = 'active ';
+			$classes[] = 'is-active ';
 		} else if ( in_array( 'current_page_item', $classes ) ) {
 			$classes = array_diff( $classes, array( 'current_page_item', 'active' ) );
-			$classes[] = 'active ';
+			$classes[] = 'is-active ';
```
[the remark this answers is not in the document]
wmcconnell: Changed. I'm confused, though. Does the second if/then clause only apply to other menus?

zainabD (approved):
Looks good!

PR #274, opened 2021-09-17
Foldable card - accordion

zainabD:
This is an ACF block for the **Foldable Card**, and can be used as an accordion if multiple cards were added into one group block:
Screen Shot 2021-09-17 at 9 23 11 AM

It includes all the available options in storybook: 
- All different colors are available.
- Card with icon.
- Collapse option.
Screen Shot 2021-09-17 at 9 26 09 AM
Screen Shot 2021-09-17 at 9 30 05 AM

To make an accordion block of this foldable card component:

1. Add multiple cards.
2. Select them all and group them into a group block (from the 3 dots at the right top corner, select "group" option).
Screen Shot 2021-09-17 at 9 34 54 AM
3. In the side panel of the group block, add "**accordion**" class to the **additional CSS classes** field, and add an id to this block from the **HTML anchor** field.
Screen Shot 2021-09-17 at 9 36 29 AM


wmcconnell:
Just wanted to comment on the small changes I made:

The button group used for selecting the highlight color of a foldable card didn't quite fit when the controls were in the sidebar:

Screen Shot 2021-09-22 at 2 50 39 PM

Since you can't resize the sidebar, I changed 'Light Gray' and 'Dark Gray' to just 'Light' and 'Dark':

Screen Shot 2021-09-22 at 2 55 19 PM

I also updated the instructions for adding an icon to make sure the user new to include the FontAwesome family ('fab', 'fas', etc.), as that is now required:

Screen Shot 2021-09-22 at 2 55 19 PM

inc/enqueue.php
```diff
@@ -62,8 +62,8 @@ function uds_wp_admin_scripts() {
 		$fa_js_version = $theme_version . '.' . filemtime( get_template_directory() . '/js/fontawesome/all.min.js' );
 		wp_enqueue_script( 'uds-wordpress-fa-scripts', get_template_directory_uri() . '/js/fontawesome/all.min.js', array(), $fa_js_version, true );
 
-		$js_version = $theme_version . '.' . filemtime( get_template_directory() . '/js/core-list-block.js' );
-		wp_enqueue_script( 'uds-wordpress-admin-scripts', get_template_directory_uri() . '/js/core-list-block.js', array(), $js_version );
+		$js_core_list_version = $theme_version . '.' . filemtime( get_template_directory() . '/js/core-list-block.js' );
```
wmcconnell, line 65: We should be careful about working in changes that are not related to the focus of the branch...presuming this is not related to foldable cards.

wmcconnell (approved):
Looks good! Glad we took the time to explore using the `<Innerblocks>` technique. It allows for a lot of flexibility within the foldable cards, like this:

Screen Shot 2021-09-22 at 2 30 29 PM

I have created both stand-alone cards and multiple accordions on the same page and have not encountered any issues - either with simple text content, or when I essentially added a miniature page layout inside a foldable card as shown above.

There is an issue with putting UDS cards inside a foldable card, but that's not anything we did - just some CSS choices from upstream that we can address in our own update if/when we need to. The foldable cards here are working as expected..

PR #275, opened 2021-09-28
UDS-Image Block

zainabD:


wmcconnell (changes requested):
Looks pretty good, but there are a few things we should change:

**Captions**
We're using default text in the caption field, which means there is **always** a caption, even if you don't want one. The logic in the code makes a caption if this field is not empty. But, since this is default text - and not a placeholder - there is always something here, and always a caption.

Screen Shot 2021-09-29 at 2 15 52 PM

It doesn't look like ACF has a way to use placeholder text here, so we should either:
- remove the default text, leaving a blank editor
- only show the text editor when the user has checked a Yes/No button to say they actually want a caption (I prefer this method)

**'Small' Image Size**
I see that we added a 25% size button based on what is in Storybook, but I think the Storybook example is really just that - an _example_ of making a smaller container if you want a smaller image. I don't think 25% is the **only** official smaller size. I don't think they were clear about that.

I think we could use a range input here to allow the user to apply different scaling options. Something like this, that could go from 25% to 100%, in steps of 5 or so:

Screen Shot 2021-09-29 at 2 35 48 PM

We could also limit it to just certain sizes with a button group: 25% 50% 75% and 100%.

Other than that, I like that we have our own image block for UDS images specifically. Still need to do some more testing in page and post contexts, but it's looking pretty good so far.

wmcconnell:
Another thought: based on [this article](https://support.advancedcustomfields.com/forums/topic/accessing-image-caption-and-description/), we should be able to get the caption that is in the media library. We're already getting our image data as an array, the caption should be in there.

The logic would be:

1. Did we get a caption with the image that was selected? If so, just use that one
2. If there was no caption in the image library, did they check the caption box? If so, check for a caption from our block and use that if available

zainabD:
Good points Walt! 
Regarding the second point, I agree with you, and if you still have that image scale field added to your local json file, please push the update to this branch. 
I liked the scale idea, and don't think that we need to limit it to certain sizes. 

---

Back to your first comment, I added the default text intentionally to make it clear to the user that this is how UDS-image looks like, because that what makes it different than the regular image.
So if the user wants to add an image with no caption, they can always use the core/image block to do so, but the main purpose of having a separate block for UDS-image is the caption style. 
So, I think it's good to show the user that by default this image has a caption with a specific css style, then they can decide whether to change the text or to remove the caption.

zainabD:
> Another thought: based on [this article](https://support.advancedcustomfields.com/forums/topic/accessing-image-caption-and-description/), we should be able to get the caption that is in the media library. We're already getting our image data as an array, the caption should be in there.
> 
> The logic would be:
> 
> 1. Did we get a caption with the image that was selected? If so, just use that one
> 2. If there was no caption in the image library, did they check the caption box? If so, check for a caption from our block and use that if available

What if the user wants to use the same image in different places with different captions?

wmcconnell:
Zainab,

- I did create an updated field group, just to make that screenshot, but then I deleted it. I'll go ahead and build it again, and see if I can get the code working as well.
- I would think that if someone knew that they wanted to use different captions, then they would simply not put one in the media library. But...if they **did** put a caption on the image in the media library, they probably meant for it to be the only caption for that image. I'll have to think about that one.

zainabD:
I updated the image caption to read the ['caption'] value first, then to check the ACF_caption field if nor ['caption'] was added as Walt suggested.

zainabD:
I added the option to wrap UDS image with text: 
Screen Shot 2021-10-08 at 5 06 57 PM

PR #276, opened 2021-09-28
Add instructions to foldable cards

zainabD:
I added a description to foldable card block to show how to make multiple foldable cards into an accordion.


wmcconnell (approved):
A good idea to add help text to our blocks when needed. I moved this text into its own tab, just to keep the basic interface for a single card as clean as possible, and got Zainab's approval on that. Approved.

PR #278, opened 2021-09-29
Tabbed Panels Block

zainabD:
In this PR I created a new ACF block for Tabbed Panels. 
Screen Shot 2021-09-29 at 3 36 39 PM

It allows to add a single tab at a time: 
Screen Shot 2021-09-29 at 3 29 05 PM

And has two fields, one for **title**, and one for **icon** to be shown next to the title: 
Screen Shot 2021-09-29 at 3 32 50 PM
Screen Shot 2021-09-29 at 3 31 27 PM

If you switch to **preview** mode, you can add any Gutenberg blocks to the tab content:
Screen Shot 2021-09-29 at 3 33 52 PM

Screen Shot 2021-09-29 at 3 34 39 PM

Then, to add multiple tabs together to show as one tabbed panel, you can select all single tabs, then group them into a group block (from the 3 dots at the right top corner, select "**group**" option):
Screen Shot 2021-09-29 at 3 41 31 PM

### DIFF
--- a/acf-json/group_615496a6a9c27.json
+++ b/acf-json/group_615496a6a9c27.json
@@ -1,0 +1,62 @@
+{
+    "key": "group_615496a6a9c27",
+    "title": "UDS Tabbed Panels",
+    "fields": [
+        {
+            "key": "field_615496a6ae434",
+            "label": "Tab title",
+            "name": "uds_single_tab_title",
+            "type": "text",
+            "instructions": "",
+            "required": 0,
+            "conditional_logic": 0,
+            "wrapper": {
+                "width": "",
+                "class": "",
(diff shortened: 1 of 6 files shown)


wmcconnell (commented):
I'm getting the left/right arrows when I hover over the tabs only when in the editor:

Screen Shot 2021-10-01 at 5 03 23 PM

And in the front-end, on mobile sizes, no arrows and a scroll bar:

Screen Shot 2021-10-01 at 5 07 40 PM

This is extra weird because `enqueue.php` looks like it's enqueuing it on the front end and **not** on the back end - which is the opposite of what I'm seeing. Did I miss something?

PR #279, opened 2021-10-01
wp-block-separator

zainabD:
Wordpress currently updated their separator block style and added two borders to it: 
Screen Shot 2021-10-01 at 10 55 58 AM

So I added this line to override the WP styling: 
```
.wp-block-separator{
	border-top:0;
	border-bottom:0;
}
```


wmcconnell (approved):
Seems like the only way to fix this while we wait for WordPress to do something. What a weird choice for them to make. Approved.

PR #303, opened 2021-12-29
Update header to UDS React and pre-stage UDS Cookie Consent

nathan-rollins:
This PR replaces the WordPress header logic in the theme with the official UDS React Header component. The UDS Header now supports customizable:

* Site logo (defaults to the standard ASU logos for desktop and mobile views)
* Partner logo (e.g. Starbucks or Mayo Clinic)
* CTA buttons in the header and in the dropdown menus
* Alternate mobile nav menu

Not all configuration settings have been enabled in this PR. This PR just enables the basic functionality required in the sites we have recently built. Example code has been placed in inc/enqueue.php to demonstrate the additional data properties that can be sent to the Header. This additional functionality will also require modifications to the theme Customizer or Menu screens.

The current React props that can or will need to be added to the theme are as follows:

- [X] loggedIn - T/F : is user signed in to WP?
- [X]  loginLink - destination url when clicking 'sign in' in header (default: /wp-admin)
- [X] logoutLink - destination url when clicking 'log out' (default: sign out of ASU CAS and WP site)
- [X] userName - user's name to be displayed in header (default: ASURITE)
- [x] navTree - formatted array of navigation menu
- [x] mobileNavTree - formatted array for alternate mobile navigation menu
- [x] expandOnHover - T/F : open dropdowns on mouse hover (default: false) - _hardcode as false_
- [ ] baseUrl - base url for all header links; important for subfolder multisites where site navigation is based at "subfolder/" instead of "/" (default: "/")
- [x] logo - array of string properties to override the primary header logo for desktop and mobile views, plus the alt text and url (default: ASU logo)
- [x] isPartner - T/F : display the approved Partner logo in the header?
- [x] partnerLogo - array of string properties to set the partner header logo (default: none)
- [X] title - Site title
- [x] animateTitle - T/F : animate the site title when first navigating to this site? (This should default to True)
- [X] parentOrg - Optional title of the parent organization
- [X] parentOrgUrl - Optional url for the parent organization
- [x] breakpoint - ["Lg", "Xl"] - at which breakpoint does the header switch between mobile and desktop layouts
- [x] buttons - array of values for generating the optional CTA buttons in the header

The UDS Header and Cookie Consent components are now installed via npm and integrated into the Gulp build scripts. The UDS Cookie Consent is currently disabled due to build issues with the upstream npm package.


jkcox:
There is some conflicting CSS I can't quite figure out yet. 

When the demo javascript menu `AsuHeader` is loaded in `enqueue.php`, there are no CSS issues. When the WP `primary` menu is loaded, there is a sass file overriding ASU styling `_reboot.scss`

jkcox:
> There is some conflicting CSS I can't quite figure out yet.
> 
> When the demo javascript menu `AsuHeader` is loaded in `enqueue.php`, there are no CSS issues. When the WP `primary` menu is loaded, there is a sass file overriding ASU styling `_reboot.scss`

This is the only issue remaining from my QA, all functionality is working as expected

SteveRyan-ASU:
Hey team. 😄 This looks awesome! 

Still digging into the proposed changes here a little bit, but I noticed some edge cases that maybe could use some extra attention? These scenarios were encountered during the initial install and configuration of the theme based on a fresh WP install and `git clone.` 

### Default menu content, lots of PHP warnings.
When I installed the theme and activated it for the first time, the theme deployed the example menu objects but it also encountered a lot of PHP warnings. (The errors may be related to some missing objects that only get set once you visit `nav-menus.php` for the first time?)
- This is really hard to replicate, but I think you can get there by switching the active theme to something else, wiping the DB and then switching it back to UDS-WordPress.
- Possible resolution: Consider eliminating the sample menu object creation script. Do you still need it?
- Why remove it? The header has a natural state that happens when there is no actual navigation content present. The default content actually makes it harder for a user to see how to achieve the "no navigation" header state. 

image

### No menu set, menu has no nav items
With the DB empty of most settings, the menu usually starts in an "unset" state with no menu object created and/or assigned to "primary" slot. When WP is in this configuration, there is no header object rendered at all. 
- Instead, I think this state should probably default to the "no navigation" state [as pictured here within Unity](https://unity.web.asu.edu/@asu-design-system/component-header/index.html?path=/story/uds-asu-header--no-navigation).
- Allows the site title and site subtitle to be displayed at all times.

image

Similar story happens when you set a menu object, but that menu has no actual nav items. Currently, in that state, the ASU Header object isn't rendered at all. It should also probably render the "no navigation" state so that the site title and subtitle are always rendered. 

image

inc/wp-custom-menu.php
```diff
@@ -8,6 +8,355 @@
...
+				}
+			}
+			// The UDS nav menu requires that we re-format our menu.
+			// We must reset the menu IDs from the array keys to sequential array keys, 0 to x.
+			// And the items[] nested arrays must be wrapped in an additional array.
+			$menu['nav-items'] = array();
+			$menu['cta-buttons'] = $cta_buttons;
+			$menu['logo-override'] = $asu_logo_override_array;
+			$menu['show-partner-logo'] = $show_partner_logo;
+			$menu['partner-logo'] = $add_partner_logo_array;
```
SteveRyan-ASU, line 259: These two lines trigger PHP warnings for undefined variables. I think they probably need to be initialized as empty arrays in line 40 and 49 before things are assigned to them? (Or, set an empty array as the `else` part of that logic?)

image

SteveRyan-ASU:
One last thing that I noticed. It looks like the new header changed the markup for the page so that there is no longer a margin-top assigned to `<main>`. As a result the content scrolls under the header and starts at the top of the browser window. 

The `_global-header.scss` SASS partial provides the needed measurements. From line 1039:
```/*--------------------------------------------------------------
8. Page Wrapper
--------------------------------------------------------------*/
// Margins are needed to prevent content from flowing under global header.

// Mobile
#asu-header + div,
#asu-header + main {
  @include transition;
  margin-top: 145px;
}

#asu-header.scrolled + div,
#asu-header.scrolled + main {
  @include transition;
  margin-top: 97px;
}

// Desktop
@include media-breakpoint-down(lg) {
  #asu-header + div,
  #asu-header + main {
    @include transition;
    margin-top: 110px;
  }
}
```

Looks like the React header is rendered with a new container wrapper + the actual ID of the rendered element is now different (#asu-header in BS4 vs #asuHeader in React). I'm thinking that this will just need to be included directly with the theme using the measurements above, but targeting the new markup.

The new markup looks like the following screen cap with the `#cookie-consent-container` div rendering between the header and the main. Moving that div elsewhere allows us to write a similar adjacent sibling CSS rule as what's captured above. Otherwise, we can probably get away with targeting `<main>` directly. 

image

If we DO align on the "right" method here, I'd be happy to submit the patch back to the Unity project so that the correct markup eventually winds up in the included SASS files (as opposed to just the theme).

inc/wp-custom-menu.php
```diff
@@ -8,6 +8,355 @@
...
+					$dropdown[ $m->menu_item_parent ]['items'][ $m->ID ] = $column[ $m->ID ];
+					
+					/**
+					 * Determine this item's top-menu item (grandparent) by getting the parent ID of this item's parent.
+					 * Adding a check here to ensure that there is a parent array in the parent of this item for us to
+					 * add anything to.
+					 */
+					if ( array_key_exists( 'parent', $dropdown[ $m->menu_item_parent ] ) ) {
+						$top_menu = $dropdown[ $m->menu_item_parent ]['parent'];
+			 /* At step 3, if menu items are still leftover, we have a column header above our menu items
```
SteveRyan-ASU, line 216: These lines here looked odd and at various points produced a lot of PHP warnings. Upon a closer inspection, I think that this may have been inadvertently pasted or included from another version of this walker script. 
- The variable `$our_array_menu` is never referenced elsewhere in the code.
- `$middle_menu` is an array here not an object.

And the most telling: If you comment this whole section out, things still appear to function normally. 😄 

The logic in line 248 is just about the same as the logic in line 231. Pretty sure that the second section here beginning at line 235 is the correct logic for getting those grandchildren into the correct place.

inc/wp-custom-menu.php
```diff
@@ -8,6 +8,355 @@
...
+				$middle_menu_column[ $middle_menu->ID ]                = array();
+			   $middle_menu_column[ $middle_menu->ID ]['ID']          = $middle_menu['ID'];
+			   $middle_menu_column[ $middle_menu->ID ]['type']        = "heading";
+			   $middle_menu_column[ $middle_menu->ID ]['text']        = $middle_menu['text'];
+			   //$middle_menu_column[ $middle_menu->ID ]['href']        = "#";
+			   $middle_menu_column[ $middle_menu->ID ]['has_current'] = false;
+
+			   $pre_menu[ $top_menu ]['items'][ $m->menu_item_parent ]['items'][ $middle_menu->ID ] = $middle_menu_column[ $middle_menu->ID ];
+
+			 }
```
SteveRyan-ASU, line 233: These lines here looked odd and at various points produced a lot of PHP warnings. Upon a closer inspection, I think that this may have been inadvertently pasted or included from another version of this walker script.

The variable $our_array_menu is never referenced elsewhere in the code.
$middle_menu is an array here not an object.
And the most telling: If you comment this whole section out, things still appear to function normally. 😄

The logic in line 248 is just about the same as the logic in line 231. Pretty sure that the second section here beginning at line 235 is the correct logic for getting those grandchildren into the correct place.

PR #320, opened 2022-05-06
core/list fix performance issues

zainabD:
We recently noticed a performance issue caused by the core/list block.
The issue was coming from the old way of registering styles to core block which were added to the js file. 
In this PR I removed all the old js code and created a theme.json file instead where I updated the core/list block from there. 
The new updates should not break the old lists created in the old way.

The text colors replace the styles in the old way: 

Screen Shot 2022-05-06 at 3 25 06 PM

To add a list with a background color, the bg-color class needs to be added to the wrapper container.
To create an order list with a step style, the class uds-steplist needs to be added to the list block: 
Screen Shot 2022-05-06 at 3 30 07 PM


nathan-rollins (approved):
Merged dev branch in and verified list color formatting. Looks good!

PR #324, opened 2022-06-02
iSearch list - border-bottom

zainabD:
updated sass file to remove the border-bottom from the last child in the isearch list


nathan-rollins:
This styling should actually be placed with the iSearch plugin. Maybe it already has? Regardless this PR should be closed and not merged in.

nathan-rollins (changes requested):
Please close this PR and merge the styling fix into iSearch plugin.

PR #337, opened 2022-08-08
Prevent Google from crawling our Sign In and Sign Out Links

wmcconnell:
We recently updated our theme to have the global 'Sign In' and 'Sign Out' links point to the local WordPress admin (/wp-admin) instead of my.asu.edu, but didn't realize that we need to prevent Google from trying to follow those links and sign in to our sites.

Not only is this a possible security issue, but signing in fires off a string of redirects - first within WordPress, then to ASU's SSO, then back - that cause errors in Google Search Console.

Changes suggested in this PR:

- Add "rel=nofollow" to both links to prevent Google from following them

### DIFF
--- a/header.php
+++ b/header.php
@@ -141,9 +141,9 @@ if (!empty($c_options['hotjar_site_id'])) {
 							<div class="nav-link login-status">
 								<?php
 								if (is_user_logged_in()) {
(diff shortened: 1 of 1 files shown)


zainabD (changes requested):
Looks good, but there are another sign in/out links down in the same file for mobile menu.
We probably need to add nofollow to these too.

zainabD (approved):

nathan-rollins (approved):

PR #350, opened 2022-10-12
core/image

zainabD:
In this PR:
- I updated the css of the core/image block to make it look like the UDS image.
- And added a JS file to unregister the core styles on core/image block and add the UDS styles instead. 

The new styles should work with all image sizes, and alignments. 

Screen Shot 2022-10-12 at 10 20 38 AM


jkcox:
What is your workflow for testing this?
I tried using Local and creating a new UDS WP theme site, then copy this branch folder to overwrite the theme in the local site.  There are some JS errors when editing, I can't see the unity blocks it only loads the default WP blocks.

`Uncaught TypeError: Cannot read properties of null (reading 'dataset')` I did `npm install` do I need to `composer install` again? Or is it easier setup using Lando?

jkcox (dismissed):
Working for me, styles, alignments, captions - no errors

PR #353, opened 2022-11-06
fix scroll behavior when click a[href^="#"]

zainabD:
In this PR I fixed the scroll behavior when clicking any `<a> `tag, by removing the **scroll to div** function that we have added previously, and adding a css smooth scroll to the html tag instead. 
- It seems like the scroll function was already added to theme.js, and we only needed that to be a smooth scroll. 
- This will also fix the foldable cards scroll behavior.


zainabD:
@jkcox Try again now, it should be fixed.

jkcox (approved):

PR #354, opened 2022-11-16
Add Maroon and Gold Icons To Cards

wmcconnell:
We have received some designs that use gold icons on standard UDS cards. This was not part of the original card specification - but seems like a logical feature to have. This pull request adds the ability to choose colors for icons when using the icon card style.

- Update the ACF fields for cards to add an 'Icon Color' option. The choices are **black**, **gold**, and **maroon**.
  - This field only shows up when the 'Icon' style has been chosen
  - To avoid conflicts with existing cards, the default color is **black**, and choosing black will not apply any new markup to the card
- Update the block rendering code to retrieve and apply the new color to the icon
  - Choosing **gold** or **maroon** will wrap the icon in a `<span>` tag that applies the appropriate color. This is a very literal check in the code: it looks for "gold" or "maroon" as the value, and only those two options will create the `<span>`.
  - I think we should stick to black/gold/maroon as the only options. If we ever do want to expand the list of available colors, we should consider flipping the logic to check for **not** black.

Here's an example from my local dev site:

Screen Shot 2022-11-16 at 2 15 36 PM


templates-blocks/cards/card.php
```diff
@@ -81,7 +85,13 @@
 		<img class="card-img-top" src="<?php echo $image_url; ?>" alt="<?php echo $image_alt; ?>">
 	<?php endif; ?>
 	<?php if ('icon' == $header_style) : ?>
-		<i class="fas fa-<?php echo $icon_name; ?> fa-2x card-icon-top"></i>
+		<?php if ('maroon' == $icon_color || 'gold' == $icon_color) : ?>
+			<span class="text-<?php echo $icon_color; ?>">
+		<?php endif; ?>
+		<span class="fas fa-<?php echo $icon_name; ?> fa-2x card-icon-top"></span>
```
zainabD, line 91: The color works with no issues. 
I just want to suggest that we change the $icon_name here to take the whole thing, like "fas fa-user" or fab fa-facebook" and so..

jkcox: Yes, agreed. There are some free icons prepended with `fab` I have run across in v4 free version used in webspark if I remember correctly, not sure if that has changed in v5 or higher.
I agree to take the whole name for compatibility with all icons, just in case.

wmcconnell: I updated the code to require the full icon name per your notes. I was worried that legacy icon cards might fail, so I put some checks in there to do the following:

- Look to see if the string starts with `fas` or `fab` as we now expect it to
- If it **does not** start with `fas` or `fab`, presume it's an old icon card with only the base icon name
- Clean it up and prepend `fas fa-` like the old code used to

Here's a screenshot showing three cards working, each with a different format for the string:
- `fab fa-apple` (the new way)
- `fa-toilet` (neither way, but possible)
- `user` (the old way)

Screen Shot 2022-12-08 at 5 57 39 PM

jkcox (approved):

PR #357, opened 2022-12-07
Add Vertical Option to Profile Block

wmcconnell:
Added the ability to have a vertical layout in our manual profile block. Here's a quick example from my local dev site:

Screen Shot 2022-12-07 at 4 47 30 PM

- Add a 'vertical layout' toggle control to the field group
- Updated the block's CSS to provide a simple `flex-direction: column` method of doing the vertical layout
- Some adjustments to padding on the image and name to better match what our React-based blocks do
- Some general cleanup of the `.scss` file to use UDS variables where I had previously used actual `px`

Note: one thing that did creep in here without me noticing is that ACF now adds a `aria-label` option to every field in the group.


jkcox:
Look good to me - the only item not pulling through are the social icons?

zainabD:
I tested the block locally, no major issues appeared, but just few things to mention: 
- When I add the block to a container, on a vertical view, the block takes the full width of the container. I think it would be better to add a max-width to have it by default go to a certain width.
- There are address fields there but not connected to anything in the block view. 
- Social media icons on mobile are not showing properly if there are 5 or more. the display flex makes them show in one line .
- When the person name links to somewhere, the style doesn't indicate that this is a link.

wmcconnell:
I have addressed Zainab's comments with the following changes:

- Set the maximum width of the vertical content area (i.e. everything but the picture) to 270px - the width of largest possible profile image. A single vertical profile placed in a wide column will now render just like it would when set in a smaller column
- Updated logic so that address fields will now display if **either** the street or city field is **not** empty. Before this, **both** fields needed to be non-empty before we would render the address
-Updated the Social Media icon SASS to use flex box classes for spacing and wrapping so that they will now properly wrap on smaller screens
- Removed SASS rules that were preventing names from looking like standard links when there is a URL provided. Names on profiles now show as maroon with underlines, like standard UDS links

wmcconnell:
After meeting with Zainab, I made some further changes:

- The larger `featured` image size is now **only** available with the horizontal layout. If you choose `vertical` as your layout, then the `featured` option is hidden in the interface
- In vertical layouts, no matter what the `featured` value is, the profile image is limited to the regular size

Unrelated to our meeting:

- I cleaned up the admin CSS to remove unnecessary rules

zainabD (approved):
I tested it locally, and it looks good, nice work!

PR #363, opened 2023-03-03
updated the button position on overlay cards per request IR330974

zainabD:
I updated the position of the button on overlay card component per Anna's request: IR330974
She asked to more the button to the right bottom corner when there is no description: 
Screenshot 2023-03-03 at 3 39 37 PM


jkcox (commented):
To clarify - applies to "UDS Overlay Card" only

Button position switching is working as expected depending on content.

Only issue I am having is the overlay card is breaking out of its container (chrome and firefox):

Screen Shot 2023-03-07 at 10 23 52 AM

This is in a 2 column layout, with content above the overlay card. If I remove all content in column it is working correctly.

Screen Shot 2023-03-07 at 10 40 42 AM

Ok it looks like the issue is having another card in the layout - the top `.card` grows to 100% height pushing the overlay card out of the container. I don't think there's anything in CSS we could do... maybe have adjustments for n+1 cards? Or am I not doing multiple cards in a column correctly... is that not an allowed design? Container > Row > Column > multiple items in column

jkcox (approved):

PR #376, opened 2023-05-10
update theme to load menu items using @asu component-header

jkcox:
Now that build process is updated with gulp, add react header.

- ACF custom fields for header customization
- `functions.php` - init class and header script
- `gulpfile.js` - bundle init header script
- `header.php` - provide ID for header to target
-  class navtree walker.php
- `enqueue.php` - enqueue asuHeader.js outside gulp bundle
- header localizer script builds menu items
- `navigation-widget.php` - update for biodesign exceptions to allow menu items to pass to navTree - probably needs more testing on another site that has the side menu/nav-widget
- `render-partials.php` - pass new header script
- `wp-custom-menu.php` - Load requested menu object and format into hierarchical array for the custom WP nav menu builders
- `/js/*` - output from gulp
- `init-uds-header.js` - init asuHeader bundled by gulp
- logos for header
- `/src/uds-asu-header/` - output from `/dist` @asu component-header


jkcox:
`jcox-gulp-react-header` is live on https://biodesign.qa.rtd.asu.edu/

zainabD:
Screenshot 2023-05-16 at 4 47 20 PM
Are these options not supported any more?

zainabD:
In multisite
Screenshot 2023-05-16 at 4 58 07 PM


Also we had this option in multisite, where we can check the box to override the subsite's menu and make the main menu show on that particular subsite.

jkcox:
I broke some ACF fields, need to revert and refactor

jkcox:
> Are these options not supported any more?

Alternate home URL - there is a `baseUrl` prop - it passes the custom prop and the custom prop is visible in react tools, but it still gets overwritten by the default "/". I can't find a storybook from ET that implements this feature, and Pitchfork isn't using it either.

Tooltip - hardcoded in the header from ET "site title" + "home page". Will pull in customized title if changed from default

jkcox:
### E&I Blog
- [ ]  gap between hero and menu
### E&I WP 
- [ ] Pages with hero - gap between hero and menu
### E&I Training
- no issues, hero works as expected 
### Skysong
- [ ] Home, Impact, About, industry, startup, careers - hero slides under header on all pages with hero
- [ ] Browse technologies - search area slides under header
- [ ] Submit innovation - top area cut off

zainabD:
It seems like the gap issue was on all websites that uses the new react header, so I updated the css in our theme. 
Biodesign has same issue but part of the hero was hidden instead of having a gap. 
It should be fixed now. 
I still need to update the header.php for: 

- [ ] Skysong
- [ ] Quantum

jehollow:
Here are my findings:

Global Futures Conference: Menu is missing and no gap when logged in as I can tell, although the ASU search is not being used on this site
RTO Site: Everything seems fine, no gaps when logged in and the menu is intact with all the pages
Economic Development: Menu is intact and all the pages are there, but there is a gap when logged in between Medium and Large screen sizes.

zainabD:
I updated these 3 child theme to fix the issues caused by react header: 
https://github.com/ASU-KE/Skysong-Innovations-Child-Theme Private
https://github.com/ASU-KE/the-quantum-collaborative-child-theme Private
https://github.com/ASU-KE/global-futures-conference-wordpress-child-theme Private

PR #378, opened 2023-05-23
Add Block Patterns to Theme

wmcconnell:
The KE Pattern Library consists of both a collection of block patterns for use directly in Gutenberg, as well as a web-based catalog of those patterns. The website is up and running, but the library isn't really complete until we add these patterns to the theme itself.

- Add the initial 43 block patterns within a `patterns` folder in the theme. This is the WordPress standard location for patterns, and all patterns here will be auto-registered by WordPress
- Update the `functions.php` file to automatically load the new `/inc/block-patterns-categories.php` file to register categories for our patterns. While Wordpress will auto register patterns themselves, you need to explicitly create pattern categories separately

There is also a semi-related commit in here that updates our `theme.json` file to allow us to use UDS spacing classes more easily. Since I was creating patterns with certain `core` blocks, I added support for spacing controls to the `group`, `list`, and `cover` blocks - and then created a new default scale that matches the class names we are used to:

spacing-controls

This is a basic implementation limited to those blocks for now. We should investigate extending this to other blocks where possible as part of updating our theme to use the more modern block theme methods WordPress is embracing.


jkcox (approved):
This should give any user the confidence to build a nice looking web page. Really cool feature!

PR #380, opened 2023-05-24
Fix Broken Photo Card Patterns

wmcconnell:
We had three photo card patterns that were broken due to an unwanted line break in the middle of the markup; this was likely the result of me searching/replacing when I updated all patterns to use a web-based image placeholder.

Interestingly, the patterns previewed just fine on the Pattern Library website, but appeared correctly broken in the Gutenberg pattern inserter. Either way, they did **not** paste into the page correctly, as the HTML was not valid.

I previewed all the rest of the patterns through the Gutenberg pattern previewer and these were the only three with any issues.

- Update the following pattern files to remove the unwanted line break:
  - Three portrait photo cards
  - Three landscape photo cards
  - Four portrait photo cards


jkcox (approved):
Looks good to me... just curious, I see some patterns start with a container, while these photo card patterns for example are just a row, that I then need to drag into a container. 

It's looks like individual items are all at the row level, while patterns of multiple items are in a container. Definitely a training item for editors, if they don't realize they're making full width designs without the container.

PR #381, opened 2023-06-05
Fix 'Scripts' Build Task

wmcconnell:
**The Problem**

It was pointed out to me that our foldable cards had stopped working in the WordPress editor. I figured something might be wrong with the Javascript we are loading in the editor, so I tried building the script locally. I noticed that I was **not** getting an `admin-bundle.js` file after the build ran.  After some investigation, it appeared that Gulp was just silently failing the task that built the new `admin-bundle.js` file. It would complete the task without any errors displayed, but no actual `admin-bundle.js` script was appearing in my `/js` directory afterwards.

In the case that you already had an `admin-bundle.js` file, it would be easy to miss the fact that it wasn't being updated. By commenting out certain lines, I was able to confirm via chaning file sizes when things were actually being included and when they were not.

The main issues is that Gulp can be a bit particular about the location (and type) of files it will allow in this script. There were two things that consistently caused this silent failure during my testing:

1. Including a file from a directory other than `paths.dev` in the admin script step. This could be worked around by hand-coding the path, rather than attempting to use the `paths.js` variable from the config file.
2. Including`theme.min.js` in the admin build, even when hand-coding the path. There was something about that file that Gulp did not like. 

We also had the issue that we were bundling FontAwesome and Bootstrap javascript files into our theme JS files, and then trying to add them **again** in the admin bundle step.

**Changes Suggested in this PR**

- Split up the current `scripts` task into two independent tasks so that each could have only one source of files:
  - `front-end-scripts` to build the `theme.js` files (one that is minified, the other full-size) and include Bootstrap and FontAwesome
  - `editor-scripts` to build the back-end `admin-bundle.js` file. This step pulls in the **un-minified** `theme.js` file from the front-end step.
- Add shortcuts to these two tasks to the `package.json` file so they can be run like our other tasks:
  - `npm run frontjs` for building the front-end file(s)
  - `npm run editorjs` for building the admin file
- Modify the current `scripts` task to run my two new tasks in serial, resulting in the same output we already expected from running `npm run scripts` and `npm run build`.

As a result, we should now have Bootstrap and FontAwesome being bundled into out `theme.min.js` and `theme.js` files. We then use `theme.js` as the foundation for the admin scripts, adding in our custom files before minifying them into the `admin-bundle.js` file.

**Testing**

If you want to confirm that I was seeing what I thought I was seeing, here are some steps:

- On a local dev site, stay on the `dev` branch and delete (or rename) the following files from `/js` folder in the theme:
  - `theme.min.js`
  - `theme.js`
  - `admin-bundle.js`
- now run `npm run scripts` (or `npm run build` if you want) to run our JS tasks. You should see a new `theme.min.js` and `theme.js`, but **not** `admin-bundle.js`. At least that is what was happening on my system.

Now, switch to my branch and:
- delete, or rename, the existing files in `/js`
- run `npm run frontjs`
- check your `/js` folder and confirm that both `theme.min.js` and `theme.js` are there
- run `npm run editorjs`
- check your `/js` folder and confirm that `admin-bundle.js` is there
- delete those files again
- run `npm run scripts` and confirm that all three files are there
- You'll know you have the right ones when they are ridiculously large. Bundling Bootstrap and FontAwesome ballooned even our minified files to over 1 megabyte! Ouch.


zainabD (approved):
I tested this locally and it worked with no issues.

PR #385, opened 2023-07-11
UDS list bugfix

jkcox:
Allows the user to use the UDS list features start value and reverse ordering. These are working in Wordpress, but the values are not passed to the UDS list styling.

I tried for a pure CSS solution, could be solved in 1 line, but unfortunately it's invalid: `counter-reset: listcounter attr(start);`
You can pass a number to `counter-reset` but not via a CSS function like `attr()`. I also considered how to update the UDS design system, we could use the default Wordpress `::marker` instead of `::before`, but there would be too much divergence if they ever did an update. 

The other issue is pseudo-elements are technically not a part of the DOM, you can read the properties after they are computed, but cannot update them directly via Javascript since they were computed after the page has completely loaded. 

The best workaround I found was using this data attributes method. We set the data needed via attributes, then use CSS content property to override the default `::before` content.

Currently restricted it to `ol.uds-list.is-style-steplist`, if there are other ordered lists that don't have that specific class, we can adjust as needed.


jkcox:
Waiting for WP 6.3 release that adds these features to the core block to move forward.

PR #386, opened 2023-09-12
on this page css fix, admin bar mobile fix

jkcox:
- Fix anchor menu CSS for Unity Blocks conversion to 2.0, removed position CSS from Unity Blocks, let the theme control menu position. 
- Fix admin bar not sticking on mobile view


zainabD (approved):
I tested it locally and it worked with no issues.

PR #398, opened 2023-10-27
add dataLayer init script

jkcox:
Added to `<head>` and set loading order as recommended by Google documentation, confirmed asu.edu is doing the same sequence: 

- Load gtag.js
- window.dataLayer init
- dataLayer.push


wmcconnell (approved):
I have tested this on a local dev site and have confirmed that it loads up and that scrolling and clicking events are, at the very least, being added dynamically to the `window.datalayer` object. No issues encountered, and it's a pretty simple updated.

I don't see any reason not to merge this in so we can test on a live/QA site where we can confirm that the data is, indeed, ending up where we want it to.

PR #405, opened 2024-02-23
header update v2.2.3

jkcox:


nathan-rollins:
I'm rusty on how this theme works, so I'm not sure what I am looking at here. Is the src/js/uds-asu-header the imported package from upstream asu-unity-stack? And then you ran `npm run build`?

jkcox:
Ok yes this tricks me every time - 

Header javascript file is loading via `enqueue.php` 
init function activates header in `js/custom/init-uds-header.js`
gulp will bundle everything in `/custom` into `theme.js`

`admin-bundle.js` is everything in `/admin`, so nothing has changed, only the minified var names.


My update process:
1. @ASU-KE/asu-unity-stack/packages/component-header `npm i`, `npm run build`
2. copy output from component-header `./dist` to  `/src` theme location
3. theme `npm run build` to re-bundle javascript if changes in other js files have been made.

wmcconnell (approved):
Reviewed and discussed during a working session. Approved.

PR #406, opened 2024-02-23
fix search.asu.edu url being passed

jkcox:
remove bad url params to properly pass URL to local site search on search.asu.edu


nathan-rollins (approved):
Awesome!

PR #408, opened 2024-03-26
asu header update v2.2.5, remove unused js files

jkcox:


wmcconnell (approved):
This has been here too long for a simple update. I pulled this branch down to my local site and did not encounter any showstoppers or loss of any functionality.

With no major issues I'm going to approve so we can move forward, but I'm open to any other opinions.

style.css
```diff
@@ -4,7 +4,7 @@ Theme URI: https://github.com/asu-ke-web-services/UDS-WordPress-Theme
 Author: Arizona State University Knowledge Enterprise
 Author URI: https://research.asu.edu
 Description: Combination of Automattic´s _s theme and Bootstrap 4, customized and enhanced for ASU Web Standards 2.0
-Version: 0.53
+Version: 0.54
```
wmcconnell, line 7: I believe we prefer not to have version bumps in our actual feature branches, but rather to bump the version on the main branch (develop, in this case) **after** the merge. This is just in case a branch languishes...like this one did...and the main branch version has already been bumped as a result of some other PR being merged.

src/js/uds-asu-header/vendor.umd.js.LICENSE.txt
wmcconnell, line 1: Interesting that we have two FontAwesome license mentions in this file. Hmmm. Wondering how this impacts how we deal with FontAwesome in the theme. Just an idle comment.

wmcconnell (approved):
I made a comment about the version bump being on the feature branch, but we don't have much else going on in the theme at the moment, so I'm going to go ahead and approve so we can merge this in a timely fashion so that it won't cause issues.

Good topic for our Friday meeting, however, and a reminder that we need to get our 1.0 out and then document - and start using - a proper process.

PR #412, opened 2024-05-14
redirect to previous page after CAS login

jkcox:
Subscribers on RTO need to be redirected to the page they logged in on. 
Might be useful for site editors as well.

1. User visits page
2. User clicks login > query param passed to CAS
3. CAS login > user redirected to wp-admin
4. `admin_init` runs to check if query param exists
5. redirect user if necessary


zainabD (approved):
Tested it locally and it worked with no issues.

PR #435, opened 2024-07-15
explicit enqueue lodash for admin script

jkcox:
add lodash to deps array


wmcconnell (changes requested):
A small nitpick, but can we change the script handle from `uds-wordpress-admin-core-image-script` to `uds-wordpress-admin-scripts`? I think we're re-using a line of code here, but this is not loading an image specific script. It will be friendlier to our future selves to make this more clear.

I tried to figure out how to do this myself in the context of the PR, since it's such a small change, but gave up.

wmcconnell (approved):
Approved for the hot fix. In the long run, of course, we can do what we want (apply the `uds-list` class to all created lists by default) via the new HTML API - or, **not do that** and just add it as a block style and let our users decide if they want a default WordPress list, or a UDS style list.

PR #469, opened 2024-09-19
Javascript admin cleanup - create core bundle

jkcox:
Many console errors from scripts that should only execute inside the editor.
Create new script bundle for editor only.


zainabD (approved):
Tested it locally, everything looks good.

PR #491, opened 2024-11-05
aria label for social media icons

jkcox:
Icons are set in wp-admin menu area with `prefix-icon name-suffix`

Get item name, regex between dashes, return icon name


wmcconnell (approved):
Approved after code review session.

PR #509, opened 2024-12-04
footer: update global links to latest

jkcox:


wmcconnell (approved):
Approved. Compared to links on [https://asu.edu](https://asu.edu) and they match. I can only presume that asu.edu is using the correct links ;)

PR #514, opened 2024-12-05
Fix accordion color choices, update user instructions

jkcox:


wmcconnell (approved):
Approved. No issues on my local with multiple cards in different states.

PR #526, opened 2025-01-21
footer - data layer attributes

jkcox:
add missing default `data-ga` attributes


jkcox:
needs init script -  doesn't register clicks with -ga, clicks are tracked without -ga

PR #530, opened 2025-02-10
UDS: Page Hero - update block to latest

jkcox:
- update block to latest UDS markup, refactor PHP logic
- fix video hero styles
- fix image hero loading

Future update: single word highlighting logic


wmcconnell (dismissed):
After realizing that my panic about broken video heroes had nothing to do with this update - and that there is only **one size** video hero allowed (large), I went back and tested more calmly:

- video heroes work as expected as long as they are large (with the caveat that we have that weird issue with the video chooser not always showing up)
- image heroes work as expected
- hero with button(s) worked - but I would avoid more than one whenever possible.

All that said...approved!

PR #542, opened 2025-02-26
Editor: Container block styles

jkcox:
- Editor max width to mimic front end page
- Containers max width to mimic front end containers


zainabD (dismissed):
Looks good, just resolve the conflict.

PR #543, opened 2025-03-03
Remove legacy cookie consent, compile font-awesome v6.7.1 JS'

jkcox:
- update gulp file
- remove from enqueue
- remove files
- compile font-awesome v6.7.1


wmcconnell (commented):
Ready to approve...but I did notice that we have  `<div id="cookieConsentContainer"></div>` in the `footer.php` file. I'm not sure how the new implementation is going to work, and whether or not it is expecting that `<div>` to be there.

Adding this, quite specifically, as a comment and not a request for changes. Just wondering what others might think. Do we leave that `<div>` in there? It's empty and won't cause any trouble...and, again, may even be needed.

jkcox:
> Ready to approve...but I did notice that we have `<div id="cookieConsentContainer"></div>` in the `footer.php` file. I'm not sure how the new implementation is going to work, and whether or not it is expecting that `<div>` to be there.
> 
> Adding this, quite specifically, as a comment and not a request for changes. Just wondering what others might think. Do we leave that `<div>` in there? It's empty and won't cause any trouble...and, again, may even be needed.

We can remove, that is for our init function

wmcconnell (approved):
Looks good to me after the removal of that `<div>`. It was nice, however temporarily, to load a site in a private window and not be pestered with that pop-up.

PR #547, opened 2025-03-20
update to component-header-footer, refactor for update 

jkcox:
- update gulp import
- remove legacy `vendor.js` and update JS in `enqueue.php`
- update init object name, build, minify
- del unused files


zainabD (approved):
Tested it locally and on E+I QA, looks good.

PR #549, opened 2025-03-24
404.php, deprecated feature: v1 hero

jkcox:
- rm v1 hero
- refactor for uds-hero
- updated depreciated function
- rm v1 hero sass file
- update theme sass


wmcconnell (changes requested):
It turns out that this update breaks our custom 404 pages. This approach uses  `WP_Query`, which is the correct thing to do as my old method has been deprecated, but the default behavior of `WP_Query` is to retrieve only **Posts**. Our custom 404 content, however, resides in **Pages** on several sites (including all 20+ Biodesign centers).

As a result, the query doesn't return any content and the actual 404 output will be empty in all cases where the user has chosen to retrieve content from a page (which, again, is how we've told everyone to do it).

It looks like the fix is simple: get the page's ID and use that, along with a slightly modified query, to retrieve the page content:

1. Change line 14 to get the ID of the custom 404 content page, instead of the title: `$custom_page_id = get_theme_mod( '404_page_id' );`
2. Modify the query on line 24 to use that Page ID: `$query = new WP_Query( array( 'page_id' => $custom_page_id ) );`

I actually have this working on my local copy, and I'm happy to push up my changes to the branch - but I honestly wasn't sure what it would do to the PR if I contributed to something I'm also assigned to review.

jkcox:
> It turns out that this update breaks our custom 404 pages. This approach uses `WP_Query`, which is the correct thing to do as my old method has been deprecated, but the default behavior of `WP_Query` is to retrieve only **Posts**. Our custom 404 content, however, resides in **Pages** on several sites (including all 20+ Biodesign centers).
> 
> As a result, the query doesn't return any content and the actual 404 output will be empty in all cases where the user has chosen to retrieve content from a page (which, again, is how we've told everyone to do it).
> 
> It looks like the fix is simple: get the page's ID and use that, along with a slightly modified query, to retrieve the page content:
> 
>     1. Change line 14 to get the ID of the custom 404 content page, instead of the title: `$custom_page_id = get_theme_mod( '404_page_id' );`
> 
>     2. Modify the query on line 24 to use that Page ID: `$query = new WP_Query( array( 'page_id' => $custom_page_id ) );`
> 
> 
> I actually have this working on my local copy, and I'm happy to push up my changes to the branch - but I honestly wasn't sure what it would do to the PR if I contributed to something I'm also assigned to review.

I think it should still allow your review, since I am the creator of the pull request?

try pushing and we can see what happens @wmcconnell

wmcconnell:
Okay, @jkcox, I pushed that change. To verify that it is working properly, do this on your local site:

- Create a new page using any name you want (I customarily use '404-not-found' as the name) on your local site
- Put some normal content in that page, using our standard Bootstrap container->row->column setup and publish the page
- Use the Customizer to set up a 'custom' 404 page, and select your new page from the drop-down list
- Try a URL that you know does not exist on your local setup and you should see the content of that page you made instead of the standard 404 stuff from our template

If you see this, then it didn't work, as this is the content from the template that should only be used when you select `default` as your 404 type in the Customizer

Screenshot 2025-04-08 at 4 23 02 PM

wmcconnell:
@jkcox: If you want, you can paste this in for the content:

```<!-- wp:wp-bootstrap-blocks/container {"isFluid":true,"marginAfter":"mb-0","className":"bg topo-black"} -->
<!-- wp:wp-bootstrap-blocks/container {"marginAfter":"mb-0"} -->
<!-- wp:wp-bootstrap-blocks/row {"template":"custom"} -->
<!-- wp:wp-bootstrap-blocks/column {"sizeXs":0} -->
<!-- wp:group {"style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"var:preset|spacing|uds-size-4","bottom":"var:preset|spacing|uds-size-4","left":"var:preset|spacing|uds-size-4","right":"var:preset|spacing|uds-size-4"}}},"textColor":"white","layout":{"type":"constrained"}} -->
<div class="wp-block-group has-white-color has-text-color has-link-color" style="padding-top:var(--wp--preset--spacing--uds-size-4);padding-right:var(--wp--preset--spacing--uds-size-4);padding-bottom:var(--wp--preset--spacing--uds-size-4);padding-left:var(--wp--preset--spacing--uds-size-4)"><!-- wp:heading -->
<h2 class="wp-block-heading"><span class="highlight-gold">Not found</span> content</h2>
<!-- /wp:heading -->

<!-- wp:paragraph {"className":"lead"} -->
<p class="lead">This is a custom 404 page, literally titled "404-not-found" that lives on this site</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
<!-- /wp:wp-bootstrap-blocks/column -->
<!-- /wp:wp-bootstrap-blocks/row -->
<!-- /wp:wp-bootstrap-blocks/container -->
<!-- /wp:wp-bootstrap-blocks/container -->```

jkcox:
@wmcconnell Looks good custom page 404 works for me with your changes

wmcconnell (approved):
Good idea to get rid of both the outdated hero markup, which wasn't strictly necessary, and my deprecated method of finding the custom content for a 404 page. With the updated content query working, we should be good.

PR #552, opened 2025-03-26
social media icons: aria label regex update 

jkcox:
- fa-square-threads - `threads`
- fa-instagram-square - `instagram`
- fa-x-twitter - `x`
- fa-youtube-square - `youtube`
- fa-linkedin - `linkedin`

### DIFF
--- a/inc/class-wp-social-media-walker.php
+++ b/inc/class-wp-social-media-walker.php
@@ -72,11 +72,9 @@ if ( ! class_exists( 'WP_Social_Media_Walker' ) ) {
 			$title = apply_filters( 'the_title', $item->title, $item->ID );
 
 			$aria_label = '';
-			preg_match('/-(?!fa-|square-|fa-square-)(.*?)(?<!-square)(?<!fa-square)-|fa-square-(.*)/', $title, $match);
+			preg_match('/fa-(?:square-)?(?:x-)?([^-]+)/', $title, $match);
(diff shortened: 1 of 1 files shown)


wmcconnell:
I noticed that there were a few Social Media icons that have multi-word icon names, and the current regex only returns the portion before the hyphen. Examples:

- `fa-square-pied-piper` returns an ARIA label of `pied`
- `fa-square-google-plus` returns an ARIA label of `google`
- `fa-square-font-awesome` returns an ARIA label of `font`
- `fa-square-x-twitter` returns an ARIA label of `x` (which is correct!)

These are edge cases, I know, and most of these are not  popular social media outlets; Pied Piper doesn't even exist! I was able to accommodate these kinds of names by using `str_replace` instead of a regex. Not very sexy, but maybe easier to see what's going on and modify as needed:

```
// Remove unwanted prefixes and suffixes from the title, leaving only the icon name.
$unwanted = array(
  'fa-',
  'square-',
  '-square',
);
$aria_label = str_replace($unwanted, '', $title);
```

Granted, with this approach what used to be `x` becomes `x-twitter`, but I didn't think it was too bad.

Any thoughts?

wmcconnell (approved):
Checked on local with all the usual social media links we expect to find, plus some new ones like Bluesy. Tested with those hyphenated ones I ran across as well. All seems to be working and every link had what looks like a sensible ARIA label.

PR #553, opened 2025-03-27
UDS Blockquote: a11y style updates

jkcox:
- update to remove italics


wmcconnell (approved):
Reviewed output and confirmed that the markup matches what's in the UI Kit here: [https://zeroheight.com/9f0b32a56/p/946b5d-blockquote](https://zeroheight.com/9f0b32a56/p/946b5d-blockquote). Didn't see any issues in either standard or image variant. Approved!

PR #555, opened 2025-04-07
Fix Issues with ACF-Based Page Hero

### LINKED ISSUE #544: UDS: Page Hero - media library video item hidden
- `display: none` on media library
- only happening on some sites
- theme version does not appear to matter, working and not working on 2 different sites running v1.4

wmcconnell:
Our recent cleanup of the ACF-based hero caused formatting issues on existing sites due to some changes in how we are expected to render heroes in certain situations and updates in how the CSS targets certain elements. The issue is really with the hero's title text, as the markup needs to change based on whether or not there is any highlighting. Specifically...

- for plain text titles, text color classes should be applied directly to the `<h1>` tag, such as `<h1 class="highlight-white">`
- for titles with background highlights, the highlight needs to be applied (as it was before) with a class applied to a `<span>` that surrounds the text content, as in: `<h1><span class="highlight-gold">Your title here</span></h1>`

We are currently applying both text-color and highlighting to the entire title, as in `<h1><span class="highlight-gold text-white">Your title here</span></h1>`, which breaks a lot of hero titles.

On top of that:
- we had removed support for highlighting a single word, or string of words, within a title; this feature was in active use on at least one site, and had been specifically requested
- the ACF based hero did not support the sub-title feature
- there was an issue with the video file picker not showing up when a video hero was selected. This appears to be fixed for me, and closes #544

- Update the hero title rendering so that the various highlight classes are in the right places (see above)
- Update the field group and rendering to support adding a sub-title with either plain, or highlighted, text
- Return the option to highlight one or more words, but not the entire title
  - Insert some inline CSS to fix the fact that the CSS will apply left margin to all `<span>` tags in the hero. We don't want this on a single-word highlight
- Several user interface updates:
  - Split up interface elements so that each part of the hero has its own tab: overall style and size, heading, sub-heading, content text, buttons
  - Put the tabs on the side and not on the top
  - Remove drop-down lists and replace with radio buttons for better clarity
  - Update the conditional logic around when certain elements are shown in an attempt to fix that missing video picker issue
  - Fix a bug that was preventing the URL option from working. Thanks, Co-Pilot!

All of this was done in `/templates-global/hero.php` and the ACF `JSON` file for maximum merge-ability! No need to rebuild any CSS, etc.


wmcconnell:
For the more visual people, some before/after images. Before...

before-1
before-2

and after...
Screenshot 2025-04-07 at 11 56 15 AM
Screenshot 2025-04-07 at 11 56 40 AM

jkcox (changes requested):
`Parse error: syntax error, unexpected '=>' (T_DOUBLE_ARROW) in /app/wp-content/themes/UDS-WordPress-Theme/templates-global/hero.php on line 18 Fatal error: Exception thrown without a stack frame in Unknown on line 0 `

unable to run this branch, crashes on front end

wmcconnell:
Well, that's not a good thing to see. I'll go check that out, and then try to figure out why I don't get that error.

jkcox:
@wmcconnell Co pilot: The issue is that the match expression is not supported in PHP versions earlier than 8.0

wmcconnell:
@jkcox Ha! I was just about to type the same thing! That was a contribution from CoPilot, as it turns out; it rewrote my `switch` statement using the new `match` expression - which I did not know even existed.

All of our current WP Engine sites are running PHP 8.2 (with one exception, newspace.asu.edu, but that site doesn't use our theme at all), and I am obviously using PHP 8 on my local environment. I guess we do have other users in other areas of ASU where I don't know their hosting situation, so they might be using 7.4.

Legitimate question though: do we want/need to support PHP 7.4 at this point? It stopped getting even security updates in 2022. Heck, even 8.2 is no longer "actively supported" (but is still getting security patches until 2027 if I'm reading [this page](https://www.php.net/supported-versions.php) correctly). 

I have no issue with reverting that code back to a `switch` statement - though the `match` statement does have some nice features. Thoughts?

jkcox:
I think i had local 7.4 set for Drupal.

jkcox (approved):
must use > PHP 8.0

PR #557, opened 2025-04-14
UDS Card: interactive card CTA

jkcox:
- make interactive card CTA optional. hide if field empty

### DIFF
--- a/acf-json/group_602edd7eb0e61.json
+++ b/acf-json/group_602edd7eb0e61.json
@@ -6,6 +6,7 @@
             "key": "field_603d25e83b9ce",
             "label": "Title and Text",
             "name": "",
+            "aria-label": "",
             "type": "accordion",
             "instructions": "",
             "required": 0,
@@ -23,6 +24,7 @@
             "key": "field_602f0c315bfc0",
             "label": "Title",
             "name": "title",
+            "aria-label": "",
             "type": "wysiwyg",
             "instructions": "For advanced HTML, use the text tab.",
             "required": 1,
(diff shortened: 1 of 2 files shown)


wmcconnell (commented):
I accidentally opened Query Monitor by clicking on the wrong thing, and noticed an `undefined variable: $heading` warning related to the interactive card. It's on line 110 of the `card.php` file:

`<a href="<?php $call_to_action_url = get_field('interactive_cta_url'); echo $call_to_action_url; ?>" class="btn btn-gold btn-sm" data-ga="call to action" data-ga-name="onclick" data-ga-event="link" data-ga-action="click" data-ga-type="<?php echo $data_ga_type = (str_contains($call_to_action_url, 'asu.edu') !== false || str_contains(substr($call_to_action_url, 0, 1), '/') !== false) ? 'internal link' : 'external link'; ?>" data-ga-region="main content" data-ga-section="<?php echo $heading; ?>"><?php $call_to_action = the_field('interactive_cta_button_text');  echo $call_to_action; ?></a>`

It's where we are setting `data-ga-section` and referring to that `$heading` variable that only exists on that line, as far as I can tell.

It's just a warning, but since it's the data layer I figure we ought to fix it. I don't know what `$heading` might have referred to - perhaps the card's title? -  so I wasn't able to offer up a solution. Do we need that specific data layer value, or can we get by with populating `data-ga-region` only?

jkcox (commented):
👍

zainabD (approved):
Looks good, just make sure to resolve conflict before merge.

PR #558, opened 2025-04-16
UDS Blockquote: fix quote icon a11y

jkcox:
https://asudev.jira.com/browse/WCMAA-430

- screen reader should ignore icon
- remove a11y title field, change role


wmcconnell (approved):
IIRC, this was to prevent needing alternate text for a completely decorative element like these giant quotes. If so, it works as advertised.

PR #580, opened 2025-06-11
remove custom list padding

jkcox:
- on mobile, padding is too large and crushes step lists into 1 word per line


zainabD (approved):
Tested it on E+I QA, looks good.

PR #586, opened 2025-07-23
@asu/unity-bootstrap-theme v1.26.5

jkcox:
Replace this text with a description of the pull request. If you are adding or enhancing functionality, briefly explain what you hope to add or improve. If you are fixing a bug, explain the original problem (and link to an Issue # if there is one). Try to keep it as concise as possible, but make sure you clearly explain the nature of the request.

- Replace these bullet list items with specific features
- Or details of bug fixes
- Think of them as the 'highlights' of the pull request
- Anything that will help someone who is reviewing the code, or searching the project history


wmcconnell (approved):
So, as John mentioned a while back, this one is tricky to test. It seems mostly to be style updates to components we don't actually use (i.e. the footer) and some subtle styling changes in the Bootstrap theme. We may see the effects of some of these when we update the Unity Blocks plugin, however.

I checked out this branch, added some potentially affect content (the blockquote/testimonial, some cards) and am basing this approval on the fact that nothing exploded in any obvious way.

PR #588, opened 2025-07-29
fix(data-ga-section): replace undefined $heading with acf get_field(`…

### LINKED ISSUE #584: card.php - interactive card - undefined variable: $heading warning
#557 

> I accidentally opened Query Monitor by clicking on the wrong thing, and noticed an undefined variable: $heading warning related to the interactive card. It's on line 110 of the card.php file:
> 
> <a href="<?php $call_to_action_url = get_field('interactive_cta_url'); echo $call_to_action_url; ?>" class="btn btn-gold btn-sm" data-ga="call to action" data-ga-name="onclick" data-ga-event="link" data-ga-action="click" data-ga-type="<?php echo $data_ga_type = (str_contains($call_to_action_url, 'asu.edu') !== false || str_contains(substr($call_to_action_url, 0, 1), '/') !== false) ? 'internal link' : 'external link'; ?>" data-ga-region="main content" data-ga-section="<?php echo $heading; ?>"><?php $call_to_action = the_field('interactive_cta_button_text');  echo $call_to_action; ?></a>
> 
> It's where we are setting data-ga-section and referring to that $heading variable that only exists on that line, as far as I can tell.
> 
> It's just a warning, but since it's the data layer I figure we ought to fix it. I don't know what $heading might have referred to - perhaps the card's title? - so I wasn't able to offer up a solution. Do we need that specific data layer value, or can we get by with populating data-ga-region only?

ramon-west-asu:
/UDS-WordPress-Theme/templates-blocks/cards/card.php

Replace
`data-ga-section="<?php echo $heading; ?>"`
with
`data-ga-section="<?php echo get_field('title'); ?>"`
to use the card's acf title for `data-ga-section`

### DIFF
--- a/templates-blocks/cards/card.php
+++ b/templates-blocks/cards/card.php
@@ -111,7 +111,7 @@ if (!empty($image_data)) {
 				<div class="hidden-details">
 					<p class="long-text"><?php the_field('body_text', false, false); ?></p>
 					<?php if (get_field('interactive_cta_url') && get_field('interactive_cta_button_text')) : ?>
-					<a href="<?php $call_to_action_url = get_field('interactive_cta_url'); echo $call_to_action_url; ?>" class="btn btn-gold btn-sm" data-ga="call to action" data-ga-name="onclick" data-ga-event="link" data-ga-action="click" data-ga-type="<?php echo $data_ga_type = (str_contains($call_to_action_url, 'asu.edu') !== false || str_contains(substr($call_to_action_url, 0, 1), '/') !== false) ? 'internal link' : 'external link'; ?>" data-ga-region="main content" data-ga-section="<?php echo get_field('title'); ?>"><?php $call_to_action = the_field('interactive_cta_button_text');  echo $call_to_action; ?></a>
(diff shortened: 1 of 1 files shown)


wmcconnell (changes requested):
When on this branch, the interactive cards do not work properly for me on the front end. They appear fine in the editor, for reasons I do not know, but I observed the following when viewing interactive cards, on Ramon's branch, on my local site, in Safari:

- A single card had multiple extra gold buttons, and none of them had any label text so they were very tiny
- When I made a second card on the same page, it ended up **inside** the first card!

I _think_ what happened is that our fix actually worked! But now, instead of just failing to find a value, we are getting some fancy quotes from our WYSIWYG editor that are breaking the HTML by not properly closing the `data-ga-region` tag and pushing a lot of closing `</div>` tags into that data attribute.

I was able to fix this by replacing `echo get_field('title')` with `the_field('title', false, false)`. By using that second `false` parameter we can request just the plain data without any extraneous formatting. That seems to work. It also removes the `<p>` tags that end up around the `<h3>` content.

For the record: `the_field()` automatically displays the retrieved field data, so it's functionally the same thing as `echo get_field()` [see the docs for get_field here](https://www.advancedcustomfields.com/resources/get_field/), so we could do the same thing with `get_field()`, but I opted for `the_field()` when doing my testing.

Screenshots:

Screenshot 2025-07-31 at 12 09 49 PM
Screenshot 2025-07-31 at 12 09 55 PM
Screenshot 2025-07-31 at 12 01 31 PM
Screenshot 2025-07-31 at 1 30 30 PM

ramon-west-asu:
Thanks Walt! updated { echo get_field('title') } to { the_field('title', false, false) }

wmcconnell (approved):
This seems to have done the trick. The cards are no longer exhibiting the earlier issue and I am able to create multiple cards without any of them ending up inside one of the others! Approved.

PR #593, opened 2025-08-14
UDS Button updates

jkcox:
- update register to json
- fix fontawesome icons not loading on front end - allow any variation of icon (user must enter full string)


wmcconnell (approved):
I like everything about this branch:

- It updates the registration to use a `block.json` file, and the more of that the better
- If fixes my dumb presumption that FontAwesome would never stray from `fas` (although I guess they do allow backwards compatability on that...for now)
- It removes some unneccesary `echo` commands. Not sure what was thinking there, but I either didn't know about, or had forgotten about, the `{$var}` syntax. Much cleaner now.
- It works! I was able to create multiple test buttons with various icons without issue

I should point out that I was not able to test with **Pro** icons because our FontAwesome update was on another branch. If we merge this in to Develop, and then merge Develop into the FontAwesome feature branch, I'll test again with Pro icons; not that I expect any trouble there.

PR #600, opened 2025-08-25
Fix errors and bugs when no menu items

jkcox:
Header menu would output "0" string when an empty menu with no menu items was set as the main menu. After trial and error it was not an issue with menu items, but the CTA button array.

- `isset` check on walker args, fixes PHP warning user sees `Attempt to read property "has_children" on null`
- set CTA array to null, fixes "0" string output 
- if there are only CTA buttons, set menu items to null. Otherwise home icon is output, and breaks UDS styling. Button should be vertically centered with site title, unless menu items exist, it will snap down to the menu items row.

### DIFF
--- a/inc/class-unity-react-header-navtree-walker.php
+++ b/inc/class-unity-react-header-navtree-walker.php
@@ -89,7 +89,7 @@ if ( ! class_exists('UDS_React_Header_Navtree') ) {
 				}
 
 				// Check for the presence of children. Add array wrapper for future depth.
-				if ( $args->walker->has_children ) {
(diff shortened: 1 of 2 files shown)


wmcconnell:
Adding this as a comment outsdie of the PR because I don't like the fact that leaving a comment as a PR review kills the review in progress...

I totally forgot that I fixed the issue on line 92 of `inc/class-unity-react-header-navtree-walker.php` on my local site and then just plain got distracted. For the record, it appeared to me that the issue was that `$args` is an `array`,and not an object. We were using `$args->walker->has_children`, but it should have been `$args['walker']->has_children`

The array item at index `walker` is the object. I'm not sure if my fix is any better. As long as we're not showing a zero when the menu is empty we should be good.

PR #680, opened 2025-11-19
Fix Issues With Horizontal Cards and Card Buttons

wmcconnell:
**Overview**
This pull request proposes changes to the current UDS Card Block to fix two issues that have hampered our site builders recently, though both are somewhat long-standing issues:

**Issue 1: Horizontal Cards are Broken**
Cards put into the horizontal orientation do not behave as expected and break the layout of any page where they are used. This is especially evident in the Safari browser, which appears to be less forgiving about misplaced closing `</div>` tags, and much more literal about what `height: 100%` means.

Here is a screenshot of what _should_ be three rows containing two horizontal cards each in Safari on the `develop` branch:

broken-horizontal-1

The constantly narrowing cards is a classic sign that there is some missing closing `</div>` tag, as each card seems to be nested within the previous one. Chrome shows a similar output, but at least closes the cards. They still stack up and get smaller as you go down the page, but they are distinct cards.

chrome-broken

From what I could see, there were a few contributing factors:

- We had an incorrect self-closing tag on one line of this file that looked like this: `<div {stuff here} />`. The `<div>` tag does not support that kind of self-closing syntax, and I don't know how any given browser will handle that.
- The logic for supporting the horizontal cards is tricky, with multiple areas of the code using conditionals to do specific things (including, most tellingly, opening and closing a `<div>` that **only** appears in horizontal cards) and we may have closed (or failed to close) that specific `<div>` in the right place
- We had included some additional markup in the code specific to horizontal cards that, after some poking around on other ASU sites, seems to be unnecessary, and may contribute to the issue. Specifically:
  - It does not appear necessary to add the `uds-card-arrangement` class, nor any of its more specific children, when rendering a single card as this code does. Those classes appear to be a method of creating card layouts in the absence of actual Bootstrap rows and columns. See the "Customized Learning" section on the [ASU Academics site home page](https://www.asu.edu/academics). If you toggle that `uds-card-arrangement` class, you'll notice that the cards stay cards, but are no longer in three "columns"

Fixes:

- Removed the self closing syntax so that is just a plain old opening `<div>`
- Removed the `uds-card-arrangement` Divs and their corresponding closing tags
- Moved the closing `</div>` tag for the horizontal-card-specific `card-content-wrapper` up so that it closes before the main card `<div>` closes.

**Issue 2: Fatal Errors When Creating Buttons**
When creating buttons for a card, it is possible to view/preview the page before the label of the button has been set. This creates a WordPress fatal error that makes it impossible to view or edit the page. Data from that page may be lost, as well, depending on how recent the last autosave was triggered. This happened twice during the build of a recent site and required manually restoring the markup from the database - with loss of recent content - and completely deleting and rebuilding the affected page.

The particular issue may not be in our code, but in ACF or WordPress. I can't be sure. Either way, it's triggered when someone creates a button in a card and then attempts to preview the page when the button "label" (aka the text of the button) has not been set. We then end up with a fatal error and a message about `attempting to access offset on type string` or something similar. It all boils down to having an NULL/empty value for that particular variable and then attempting to access it in order to render the button.

Oddly, this does not happen when creating a standard link using the same process and ACF component. It only seems to be an issue with buttons - and specifically with that "label" value missing. If you provide a label, but fail to provide a URL, the button will render just fine.

Fixes:
 - I modified the card code to provide a default label for buttons. To be clear, I modified the actual rendering code in `card.php` and **did not** set a default value in the block's field group, which should also fix this. When a button is rendered, we run a check to make sure that the value is set, and not empty. If those tests fail, the code will provide a button label of "Button label missing!" to alert the person editing the page that the button is not complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,35 @@
 
 ### Features
 
-- ASU Web Standards 2.0
-  - Based on ASU's official design token libray for consistent adherence to Web Standards 2.0
+- **ASU Web Standards 2.0**
+  - Based on ASU's official design token library for consistent adherence to Web Standards 2.0
   - Utilizes the official ASU Bootstrap theme built on top of the design token library
-  - Accessibilty compliant
+  - Accessibility compliant
+  - React-based header and footer components
 - WordPress
   - Utilizes standard WordPress features, such as page templates, widgets, and shortcodes, for easy and rapid development of standards-compliant WordPress sites
 - PHP 8 required
   - PHP 8 - v1.6 and later
   - PHP 7.4 compatible - v1.5 and earlier
+
+- **Content Components**
+  - 16 custom ACF blocks for UDS-compliant content (cards, banners, alerts, modals, etc.)
+  - 42+ pre-designed block patterns for rapid page building
+  - Custom widgets for global banners and footer content
+  - Shortcode for sidebar menus
+
+- **WordPress Integration**
+  - Utilizes standard WordPress features for easy development
+  - Enhanced WP REST API with featured image and ACF field support
+  - Custom navigation menus with ASU Web Standards enforcement
+  - Widget areas for banners and footer customization
+  - Child theme support with starter template available
+
+- **Analytics & Performance**
+  - Built-in Google Analytics Data Layer for comprehensive event tracking
+  - Google Tag Manager, Google Analytics, and Hotjar integration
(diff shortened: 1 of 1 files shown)


jkcox:
@wmcconnell I put these inside a container > row, not sure why they stretch off the page. They also slide under other elements

Screenshot 2025-11-24 at 12 08 17 PM

PR #721, opened 2026-02-09
alerts - remove focus function

### LINKED ISSUE #719: alert focus jumps
**Describe the bug**
dismissing alert causes focus to jump, if no element to focus it jumps far around the page.
**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

AB#43962

jkcox:
## Description

removed function calls that auto focus on the next element
## Motivation and Context

with an alert on the page, if an alert is dismissed the page will jump far to the next focus element. Let the user control focus.
## How has this been tested?

local test
n/a
## Types of changes

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/src/js/uds/data-layer.js
+++ b/src/js/uds/data-layer.js
@@ -169,16 +169,8 @@ function initDataLayer() {
 	// Alerts and banners. Events emitted by the .alert element which uses BS5 collapse.
 	document.querySelectorAll('.alert').forEach((element) => {
 
-		// While we are here, let's move the focus appropriately. We'll need the index later.
-		// Recommendation from BS 5: https://getbootstrap.com/docs/5.2/components/alerts/#dismissing
-		var allFocusableElements = document.querySelectorAll(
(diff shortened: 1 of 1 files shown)


wmcconnell (approved):
I like that this solution at least leaves things in the hands of the user. In cases like we saw when testing, it is possible that the next focusable element may be a good distance down the page, causing a lengthy, automatic scroll that will confuse users both with accessibility needs as well as those without.

In my testing, the TAB key moved to the next focusable element on the page after the alert was removed in Safari, Chrome, and Firefox. While I'm not sure that it's better to not have a focused element, we are at least removing the side effects we were seeing of seemingly random scrolling.

I would not be surprised at all if we end up having to revisit this, but for now I'll approve it since it solves the issue we were encountering.

I saw one possible solution online: replace the element being closed with a new, invisible element (still with focus) that fully removes itself `onBlur` after the user has moved focus forward or backwards themselves.

PR #728, opened 2026-02-16
UDS Banner updates

jkcox:
## Description

- FA7 icon name updates 
-  update UDS markup
- update icon markup for a11y
- use BS5 dismiss event  
## Motivation and Context

- a11y JIRA ticket
## How has this been tested?

locally
## Types of changes

## Checklist:

- [x] All new and existing tests passed.

### DIFF
--- a/templates-blocks/banner/banner.php
+++ b/templates-blocks/banner/banner.php
@@ -58,55 +58,48 @@ if ( isset( $block['className'] ) && ! empty( $block['className'] ) ) {
 
 // we'll build a 'button block' based on those conditions.
 $button_block = '';
-$button_block_open = '<div class="banner-buttons">';
-$button_block_close = '</div>';
-
-// if our button count is more than zero, we have buttons to build.
 if ( $button_count > 0 ) {
-
-	// add the opening markup to our button block.
-	$button_block .= $button_block_open;
-
-	// set a button class to make sure we don't make black buttons on a black BG.
-	// all banner colors, besides black, use the dark buttons.
-	$button_class = 'dark';
-	if ( 'black' == $color ) {
-		$button_class = 'gold';
-	}
-
-	// if we got here we have more than zero buttons, so we will render button #1.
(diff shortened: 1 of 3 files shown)


templates-blocks/banner/banner.php
```diff
@@ -58,55 +58,48 @@
...
-							<button type="button" class="btn btn-circle btn-circle-alt-white close" aria-label="Close" data-ga="banner close" data-ga-name="onclick" data-ga-event="button" data-ga-action="click" data-ga-type="close" data-ga-region="main content" data-ga-section="banner" onclick="event.target.parentNode.parentNode.style.display='none';">x</button>
-						</div>
-					<?php endif; ?>
-				</div>
+<div class="uds-full-width">
+	<div role="alert" class="banner-<?php echo $color; ?> alert alert-dismissable <?php echo $additional_classes; ?>">
+		<div class="banner uds-content-align">
+			<div class="banner-icon"><span class="fa-solid fa-<?php echo $icon; ?>" aria-hidden="true" focusable="false"></span></div>
+			<div class="banner-content">
+				<h1 tabindex="0"><?php echo $banner_title; ?></h1>
```
wmcconnell: Just wanted to make a comment here, not as part of the review. I don't understand why the markup from UDS uses an `h1` tag here; we should only have one `h1` per page, and lots of pages use a Hero and so will have an `h1` there. I suppose it was because the banner is meant to appear before the hero, but still.

Also, I don't know if it's proper accessibility to use `tabindex=0` on a non-interactive element and then not give it an `aria-role` or any interactivity. If you can focus on an element, I believe you should be able to interact via spacebar/return key.

We're using the official markup and all, but I'm not sure I agree with it.

wmcconnell (approved):
Tested locally on a page I made with multiple banners before switching to this branch. No issues arose with my existing banners, or creating new banners.

There was one change that we should be aware of:

- the new markup uses both `alert` and `alert-dismissible` classes where before we just had `banner-{color}` as the classes here.
- As a result, the banners have picked up a 16px bottom margin from the `alert` class and the corresponding  `--bs-alert-margin-bottom` CSS variable.
- Our main banner widget area is above the hero and this change will result in 16px of white space between a banner and the page hero. We will likely hear about this from our designers or clients.
- It is the official markup, however, so I don't know what to say to that

I'm approving this with that slight reservation because it's the official markup, and the data layer attributes are important.

PR #733, opened 2026-02-23
Anchor menu theme updates

jkcox:
## Description

- rm local block uds init
- allow uds js globally
## Motivation and Context

updating to latest uds functionality has been missing
## How has this been tested?

local
## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] All new and existing tests passed.


wmcconnell (approved):
This seems reasonable, and does not cause any obvious issues on my local site (though that site does not contain any anchor menus, and its state can't always be trusted). Looks like it's doing what is claimed: universally enqueuing a Bootstrap JS file and removing a specific enqueuing of it in the parallax block - as that would no longer be necessary.

PR #769, opened 2026-04-10
Fix #658 - Created a custom process stage block

### LINKED ISSUE #658: Replace Process with custom block
Image

Display stages of a process with a visually appealing animation.
### features:
- allow vertical or horizontal layout
- toggle animations
- decide on styling adjustments to allow
- process items should have: image, title, content

AB#43836

### LINKED ISSUE #785: UDS Block: Process Stages - allow user customizations
- [x] Allow users to customize number display. Make the numbers an editable text field. Provide default values that count up if user doesn't edit.
- [x] Allow users to adjust number text color (there is a built in WP style tab for this, I can show you if you need guidance)
- [x] Pause button for animations (see text marquee block for example) only display if animation selected.

We will have designers review the block, they may have a few additional suggestions (possibly slowing down animations). 

AB#44520

pateldivyesh1323:
## Description
Added a custom process block
- Added horizontal and vertical layout
- Added animation: none, bounce, pulse

## Motivation and Context
#658 - Replace process with custom block

## How has this been tested?
- Resizing the window recalculates the lines
- Changing the layout in editing mode
- Layout changes to vertical on smaller screens
- Adding multiple blocks on the same page
- Adding a stage without image

image

image

## Types of changes

## Checklist:
- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
   - Utilizes standard WordPress features, such as page templates, widgets, and shortcodes, for easy and rapid development of standards-compliant WordPress sites
   - PHP 8 required (v1.6+) / PHP 7.4 compatible (v1.5 and earlier)
 - **Content Components**
-  - 17 custom ACF blocks for UDS-compliant content (cards, banners, alerts, modals, text marquee, etc.)
+  - 19 custom ACF blocks for UDS-compliant content (cards, banners, alerts, modals, text marquee, etc.)
   - 42+ pre-designed block patterns for rapid page building
   - Custom widgets for global banners and footer content
   - Shortcode for sidebar menus
--- a/acf-json/group_69d6aa54a7fe4.json
+++ b/acf-json/group_69d6aa54a7fe4.json
@@ -3,27 +3,6 @@
     "title": "Block: Process Stages",
     "fields": [
         {
-            "key": "field_69d6aa542a687",
-            "label": "Process title",
-            "name": "process_title",
-            "aria-label": "",
-            "type": "text",
-            "instructions": "",
-            "required": 0,
-            "conditional_logic": 0,
-            "wrapper": {
-                "width": "",
-                "class": "",
-                "id": ""
-            },
-            "default_value": "Our Working Process",
-            "maxlength": "",
-            "allow_in_bindings": 0,
-            "placeholder": "",
-            "prepend": "",
-            "append": ""
-        },
-        {
             "key": "field_69d6b181715aa",
             "label": "Horizontal Layout",
             "name": "uds_layout_horizontal",
@@ -191,5 +170,5 @@
     "display_title": "",
     "allow_ai_access": false,
     "ai_description": "",
-    "modified": 1775849529
+    "modified": 1776098064
 }
--- a/acf-json/group_69d6aa54a7fe4.json
+++ b/acf-json/group_69d6aa54a7fe4.json
@@ -138,6 +138,36 @@
                     "parent_repeater": "field_69d6b1d2715ab"
                 }
             ]
+        },
+        {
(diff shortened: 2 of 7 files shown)


jkcox:
@copilot check for WCAG 2.1 Level AA compliance

jkcox (dismissed):
This is great! Tested locally.

- Redraws on resize
- Resizes to vertical on mobile
- Works with 3, 4, 5 items
- Really long content does not overlap with the next item

I added some features for accessibility requirements:

- aria attributes where needed
- `reduce-motion` CSS

I also added some metadata:
- block versioning to track changes
- block API to use new style editor

We may want to allow users to change the white font numbers to dark, depending on the images they use, or possibly add some drop shadow. Overall great work!

wmcconnell (commented):
First of all, @pateldivyesh1323, good work! I love it so far.

Based on my experience with the Qi block we are replacing, and from past discussions with designers, there were a few things I noticed that I wanted to bring up. I should note that these are from some initial functional testing, and I have not read through the actual code yet.

- The addition of a period after each number, as in `1.` instead of `1`, was something that Qi blocks did without asking, and I recall at least one occasion where we were asked if it could be removed. I would default to just the number with no period if that's possible.
- Better yet, could we add a field to the current repeater that allows users to provide the text themselves? I can picture a situation where someone would use this block with letters instead of numbers (maybe spelling out an acronym or something like that)
- Having the only option be white text may be difficult on top of many images. Below is a screenshot from my testing where I wasn't even thinking about this and ended up with three out of four images where white text does not work well:

Image

- I don't know how to handle that, to be honest. User selectable text colors might help, but it's often difficult to pick one color that works with both light and dark images. The solution to this could just be to make sure our designers provide images that work with white numbers. I just wanted to bring this up.
- I like the addition of animation, but I feel like the current values have things moving a little too fast. I would recommend slowing those animations down so they're more subtle - half as fast as they are now at the very least.
- Speaking of that, if we have animations we will need a pause button at the very least for accessibility reasons. Since @jkcox has already done that for the marquee block, that might be a good place to look for implementation details. Granted, it's always possible to just not turn on the animations, but if we add that feature I know someone will use it.

I don't want to add much more, because I'll just keep requesting features 😉 I will give a demonstration of this in our designer/dev meeting on Friday to see what input our designers have - especially around animation.

But the big take away is good work!

jkcox:
@wmcconnell if we have time, let's review all the remaining blocks with designers. Meryl is aware of what is left to build, but we should confirm we've added the requirements of each block that they would expect.

For example - we probably should have at the start of development on this block asked for user editable numbers, change font colors, etc.

Not too difficult on this one to refactor to add those features, but it just helps save development time.

mjurado22 (dismissed):
Took a pass at testing this locally. Confirmed the redraw behavior is working when switching layouts and resizing and the overflow issue looks good in horizontal layout even with longer content.

I did notice in vertical layout (page view) that with longer titles/descriptions, the connector line between stages can overlap or get very close to the text, especially on desktop. On mobile it doesn't overlap but does get tight to the last line.

Not a blocker but something to be aware of as an edge case with longer content.

Overall this looks good to me, good job!

jkcox:
Yes I double checked WCAG 2.1 - unfortunately reduce-motion CSS is not enough to qualify, we also need a pause button for users that don't toggle on reduce motion on their machines.

pateldivyesh1323:
https://github.com/user-attachments/assets/eb5b7882-ecdc-493a-a945-21e27250b44d


- Added custom text on image
- Added color picker for image text
- Slowed down animations
- Added pause button to animation

jkcox (approved):

PR #778, opened 2026-04-14
UDS Card - add missing metadata to fix data loss

### LINKED ISSUE #777: UDS Card - done button not working
**Describe the bug**
If a user fills a field using the new acf modal, clicking done loses data. User has to focus on another field to update.

**To Reproduce**
1. Add card to page
2. Click on expand editor
3. Type in a field
4. Click done

**Expected behavior**
Clicking done should preserve data.

AB#44501

jkcox:
## Description

Added missing block api version, and version to track block changes.
## Motivation and Context

Fix for done button bug. See issue
## How has this been tested?

Tested locally
## Types of changes

## Checklist:

### DIFF
--- a/templates-blocks/cards/block.json
+++ b/templates-blocks/cards/block.json
@@ -1,6 +1,8 @@
 {
     "name": "acf/uds-card",
     "title": "UDS Card",
+	"apiVersion": 3,
+	"version": "1.0.0",
     "description": "A block for building static versions of basic, degree, story, event, and interactive cards.",
     "category": "uds",
     "icon": "id-alt",


mjurado22 (changes requested):
Tested locally against the issue steps and was still able to reproduce an issue when updating only a single field in the UDS Card.

In my testing, if I changed just the title or just the body text and then saved/viewed the page, the update did not save on the first try. When returning to the editor tab, WordPress still showed unsaved changes. Saving a second time caused the change to stick.

I think my earlier tests passed because I was updating more than one card field at once, which may have forced the block to register the change. Based on this, it doesn't look like the issue is fully resolved yet.

jkcox:
Interesting, I wonder what the difference in our environments are. 

In my testing with this change, the card block updates as the user types, and it doesn't wait for blur event or the done button click. The expected Gutenberg/React experience

mjurado22:
Thanks for checking. I re-tested on a brand new page with a single UDS Card and was able to reproduce it again. I changed only the title, clicked save and the view page tab still showed the default title. When I tried to close the editor tab it warned there were unsaved changes. After clicking save a second time, the change saved correctly.

I am seeing some local console warnings on my end, so I'll take a look at that as well but with this flow I can reproduce the issue consistently. It may also be helpful to get another set of eyes on it.

wmcconnell:
Just chiming in to say that I am still not able to replicate this on a local site and, in one case, could not make it happen on the Global Futures QA site where Cindy first reported it.

I created and edited cards on both the `main` and `jcox-cardapifix` branches in all three browsers I have installed: Safari, Firefox, and Chrome. While there was always a slight, and seemingly intentional, pause after typing before text would update, it never failed to update.

Thoughts in my head that I don't want to forget:

- When Cindy was demonstrating this, I swear I briefly saw the Basic Auth login box pop up while she was on her way to the page. Could the it be auth/login based? With edits taking place via the API, I wonder if there's something failing there? But why can Melisa replicate this and not me?
- Cindy also claims that the "Open Expanded Editor" button is not showing up for her. That's been in place since v6.6 of ACF, and Global Futures QA is already on 6.8.x. Why would she not see that?
- The 'Done' button on the slide-out editing panel will animate when things are being saved. It's diagonal stripes moving, like a barber pole. I can see it animate as I'm typing. @mjurado22, can you try again and see if that button has the animated stripes when you are making edits to a card with the slide-out drawer?

Just some thoughts. I'll return to this tomorrow if there's time.

jkcox:
Yes, I get the barber pole behavior as I edit as well.

jkcox:
I think it might be tinymce/wysiwyg editor

I updated the alert block to v3, and it works fine as I would expect. When back to edit, but the text failed as Melissa described once. I tested again and got the barber pole behavior. There are warnings in console about tinymce having deprecated features.

mjurado22:
Quick follow-up from more testing on my end: in my environment, changes made in expanded editor appear to stay in the expanded editor only. When I click Done, they do not reflect back in the main editor and saving after that still does not apply them to the page.

I also don't see the barber pole animation while typing, so it doesn't appear to be updating state as I type like you described.

I have also been able to reproduce the save issue when editing directly in the main editor, where the first save does not take and the second save does.

For reference, I'm testing locally (DDEV) in Chrome. Happy to provide WP/ACF versions or test in another browser if that helps narrow it down.

jkcox:
99% sure it is tinyMCE

Cards uses tinyMCE, and I am getting that behavior that fails.

I just updated UDS Headings (or any of the other blocks I've updated so far), and it uses a simple text field. Updates immediately with the barber pole behavior.

jkcox:
Modals has a tinyMCE-like editor - maybe the built in WP one? Will need to look up how to switch/make sure we're using the same editors across all blocks.

The Modals editor works as expected updating instantly.

mjurado22:
Quick update from more testing: I found that the issue seems tide to the Visual editor mode for the title/body fields.

When I make changes in the Visual tab in the expanded editor, they do not reflect back in the main editor for me. But when I switch to the Text tab and make the same kind of change there, the update does reflect in the main editor and saves correctly.

I also tested adding a button in the expanded editor and that synced and saved correctly as well.

So on my side, it looks like the problem is more specific to the Visual/TinyMCE editor behavior for those text fields. I still do not see the barber pole animation while typing.

wmcconnell:
@jkcox @mjurado22 I think I may be on to something, but could use confirmation from one or both of you.

First, the setup:

- When I got back from my appointment today, I suddenly had the same issue you guys described...the one I was not able to replicate on Friday. Maybe something updated over the weekend? I'm not sure. Regardless, I was unable to edit card body and title using the slide-over v3 editor panel
- I saw the conversation between the two of you regarding cards not updating. I also noticed one of you had tried modals and did not have that issue and that it might be related to the WYSIWYG field
- That got me thinking: the UDS Card block has two WYSIWYG fields for title and body
  - Both of those WYSIWYG fields are set to initialize the TinyMCE editor **after** the user clicks inside the field. This is an ACF setting that can be changed by toggling a switch on the 'Presentation' tab when editing the field group.
 - Thinking that the issue might be related to attempting to initialize the WYSIWYG, I synced up the card field group on my local and (while on John's card-fix branch):
  - Changed the Title field to initialize immediately and **not wait** until I clicked in it
  - Attempted to edit the title, and it started working again!
  - Atttempted to edit the body text (still set to delay initialization) and it **did not work**
  - Updated the field definition for the body text  to initialize right away and **it worked!**
  - Changed body text back to delayed initialization and tried to edit the same card again: **did not work**

So it seems that the issue might be delaying the initialization of the TinyMCE editor. Could one of you try the same trick and see if it works? The steps would be:

- Go to your site Dashboard -> ACF -> Field Groups
- If your local set-up is "correct," you should see an empty screen (no field groups should be synced to your database)
- Click on the 'Sync Available' tab, find the 'UDS Card' group, hover your mouse over that entry and you should see - in the right-hand 'Local JSON' column - a link that says 'Import'
- Click on that link to import the card settings to your database.
- The UDS Card field group should appear on the "All" tab of the interface now and be available for you to edit
- Edit the UDS Card field group to update the Title or Body Text fields to not delay initialization:
  - Click 'Edit' on the field group
  - Click on the 'Presentation' tab and toggle the 'Delay Initialization' switch to OFF
  - Save your changes
- See if the field(s) you updated work as expected now.
- If so, that might just be culprit and we would want to set both WYSIWYG fields to immediate activation

mjurado22 (approved):
Re-tested locally after the latest update. Confirmed the expanded editor is now syncing correctly for title and body text (visual and text) and changes are reflected in the main editor and saved on the first attempt. Also verified link updates behave as expected.

Looks good to approve.

PR #787, opened 2026-04-22
Custom Block: Radial Progress

### LINKED ISSUE #655: Replace Radial Progress Bar with custom block
- animated circular progress percentage. animate from 0 -> x %
### features needed:
- ending number
- animation duration
- title
- text

### styling
possible options:
- colors for x% and unused %
- circle fill
- circle size
- line thickness for x% and unused %
- allow text styling?

[AB#43375](https://asuteam404.visualstudio.com/3c3feac4-94b8-464c-8938-c46f8a83b297/_workitems/edit/43375)

pateldivyesh1323:
## Description
Animated circular progress block.

## Motivation and Context
Issue: #655 

### Supports
- End value
- Suffix Text
- Animation
- Title
- Text
- Progress/Fill/Track/Text color
- Size
- Progress/Track Stroke width

## How has this been tested?

- Adding value greater than 100
- Changing color values
- Changing the size of block and tracks
- Different screen sizes
- Changing duration of animation

image

https://github.com/user-attachments/assets/df0657d6-acf9-4b3c-8511-c0cf25e3c9d8

image

image

## Types of changes

## Checklist:
- [x] I have updated the documentation accordingly.
- [x] (if appropriate) Run Copilot check for WCAG 2.1 Level AA compliance


jkcox (changes requested):
Looks great!

Only 1 major issue: fill color doesn't seem to be working. Looks like it might be getting overwritten or converted to a different value type.

pateldivyesh1323:
It is happening because the fill color supports transparency and is in the form of rgba. So the sanitize_hex_color function is returning an empty string because it cannot detect a hex string, and it is getting defaulted to transparent.

I can simply remove the transparency and by default keep the fill color white.
Or else I can write a function which can sanitize the rgba as well.

jkcox:
> It is happening because the fill color supports transparency and is in the form of rgba. So the sanitize_hex_color function is returning an empty string because it cannot detect a hex string, and it is getting defaulted to transparent.
> 
> I can simply remove the transparency and by default keep the fill color white. Or else I can write a function which can sanitize the rgba as well.

Let's just remove it and keep it white for now. If designers request that feature later, we can refactor.

jkcox (approved):
able to select fill color, looks good!

PR #792, opened 2026-04-27
Custom Block: Pie and Doughnut Chart

### LINKED ISSUE #659: Replace Pie and Doughnut Charts with custom block
Ideally it would be best long term to use a library like d3.js. Possibly combine this task with the bar charts, and explore options for an overall data visualization plugin?

[AB#43379](https://asuteam404.visualstudio.com/3c3feac4-94b8-464c-8938-c46f8a83b297/_workitems/edit/43379)

pateldivyesh1323:
## Description
Added Pie and Doughnut chart

## Motivation and Context
Fixes #659 

### Supports

- Chart types: Doughnut, Pie
- Animation: none, grow, rotate
- Enable/Disable legend and percentages
- Different Sizes

## How has this been tested?
- Compresses on different screen sizes
- Changing animation
- Thickness only appears on chart type: Doughnut

image

--

image

--

image

image

## Types of changes

## Checklist:
- [x] (if appropriate) Run Copilot check for WCAG 2.1 Level AA compliance


jkcox:
@copilot+claude-opus-4.6 `tab index=0` is not working. I tab through the page, and tabbing selects the same `<circle>` segment, if there are 2 segments it tabs the same segment twice. the user would hear the same aria-label read twice. any suggestions?

jkcox:
@copilot We'd like the user to be able to tab through the `<circle>` items that have `aria-label` so they get read correctly, but the user should not be able to tab through the chart legend. Or vice versa if that is easier.

In the chart legend, items have `tabindex=0` but are still able to tab through items. same with `<circle>` items that have `tabindex=0` it doesn't work.

Users should be able to tab through either the legend or the chart, and get the right `aria-label` reading

jkcox (approved):
Nice job.

Fixed issue with tabbing through items, to avoid screen readers reading things twice.

I like the hover states on the chart, and the connection of the hover states between the chart and the legend.

